PR #414, opened 2017-03-29
412-invalid unknown effect var_p

icebert:
Fix invalid unknown effect var_p when var_p.posedit.uncertain=False.


reece:
Thanks for the PR, Meng.

I'm wondering about a different approach and would like your input.

This issue is really about how we represent the notion of an unknown protein effect. Your PR does this with special values for `posedit.pos` and `posedit.edit`. As an alternative, we could use var.posedit = None to indicate unknown effect.

Throughout the code, `pos` and `edit` are typically represented with class instances. We use None in some cases, which seems reasonable. My concern with the existing code is that we use some special case scalar values for `pos` and `edit` in places. For example, we use `edit == "0"` to indicate "no protein" (https://github.com/biocommons/hgvs/blob/master/hgvs/utils/altseq_to_hgvsp.py#L318).

Your PR continues the use of special values and is consistent with current code. 

However, I wonder whether we should move away from special values and toward explicit representation of these states.

As a practical example of the significance, suppose a user wants to know whether the protein effect is uncertain. With this PR, they would need to reimplement the logic regarding `pos` and `edit`. With the proposed alternative, they would use `var.posedit is None`, which I think is more straightforward.

I think two options exist:
* Update this PR to include a method like "protein_effect_unknown" which centralizes the logic
* Use `posedit = None` to represent unknown effect.

What do you think?

reece:
@lawrlee You have a good design eye and have used hgvs. Do you have comment on this issue?

lawrlee:
I would prefer a single conditional check like `posedit == None` rather than two pieces of logic on `pos` and `edit`. If I recall correctly, the `SequenceVariant` object does not contain property-like attributes like `protein_effect_unknown`, so that would be a departure from current norms.

icebert:
Using 'posedit = None' is a good idea. This will help to make the concept and logic to be simple.

reece:
Dropped in favor of PR #416. Thanks Meng!

PR #423, opened 2017-04-12
update URL in CHANGELOG

chenliangomc:


reece:
Thanks for the fix, @chenliangomc.

PR #427, opened 2017-04-22
426-Ensure c. coordinate within CDS bound

icebert:
1. Add validation of c coordinate which should be within CDS bound in ExtrinsicValidator

2. Incorporating configurable extrinsic validate before variant mapping


reece:
I wonder about generalizing `ext_validate` to something like `prevalidation_level`, which could be an ordered enum of [None, Intrinsic, Extrinsic]. This would permit callers to bypass some or all of the pre-mapping validation if they know the variant is already valid.

What do you think?

icebert:
Good idea! I'll update this pull request.

icebert:
Hi Reece, I just implement the idea of prevalidation_level. All tests passed.

Issue #424, opened 2017-04-12
stringified hgvs variants are not using the c.BASE(s)= nomenclature when using 1.0.1post3


PeteCausey-Freeman:
Hi Reece and Meng

The previous version of hgvs that I used (1.0.0a1) correctly displayed variants where posedit.edit.alt was equivalent to posedit.edit.ref as Reference:type.PositionBases= e.g. NC_000017.11:g.50198002C= or NM_000088.3:c.589G=.

Certainly with respect to genomic variants (I have not tested transcript variants, 0.1.0post1 has reverted back to using NC_000017.11:g.50198002= . Please can you put behaviour back to the way it was in the 1.0.0a1 release, particularly because I believe that HGVS guidelines would recommend that the reference bases are displayed.

Thanks

Pete

reece:
Sorry about this, Pete. I'm flabbergasted -- I was nearly certain that we had a test for this when we addressed it previously.

PeteCausey-Freeman:
No problem. 

Thanks 
________________________________________

icebert:
I recall what causes this. I added the max_ref_length in formatter, which was set to be 0 by default. This leads to the missing of ref in ident variants. I just realize that this should not be applied to ident variant and I'll fix this.

reece:
@PeteCausey-Freeman , @meng: I'd like to discuss how this should work. The upshot is that I think we should leave the code as-is, but I want your opinions. (Sorry Meng -- I see that you already submitted a PR.)

First, the cause (for Peter): We recently added a configuration parameter, max_ref_length, that limits the size of the deleted reference sequence displayed for del and delins. Since the deleted sequence is always implied by the reference, it's always legal to omit it. Adding the deleted sequence is particularly useful for relatively small changes. That is, we're happy to look at delACGinsTAGCAC, but less happy if the del sequence is long. We made the default max_ref_length = 0, so that the del sequence is not shown by default. This issue for identity variants occurs because the stringification for identity variants uses the same reference sequence length. With the current code, you can do this:

```Python
>>> v = hp.parse_hgvs_variant("NC_000017.11:g.50198002C=")

# by default:
>>> v.format()
u'NC_000017.11:g.50198002='

# one-time format with custom config:
>>> v.format(conf={"max_ref_length": 10})
u'NC_000017.11:g.50198002C='

# change the global setting
>>> hgvs.global_config.formatting.max_ref_length = 5
>>> v.format()
u'NC_000017.11:g.50198002C='
[...]

PR #429, opened 2017-04-26
424-Include ref seq in ident variant

icebert:
The reference sequence in ident variants got loss because of 0 max_ref_length in formatter.


reece:
Closed without merging per discussion in #424.

PR #446, opened 2017-05-28
445-fix start is 0 when var is normalized to be inserted before the start of transcript

icebert:
Ensure start is not 0 after normalizing. All tests pass.


hgvs/normalizer.py
```diff
@@ -155,6 +155,14 @@ def normalize(self, var):
                 ref_end = end - 1
                 edit = hgvs.edit.NARefAlt(ref=ref, alt=alt)
 
+        # ensure the start is not 0
+        if ref_start == 0:
```
reece: Is there an analogous issue on the 3' end that also needs a special case?

reece:
This is a great catch, Meng. Thanks for the PR. I added a comment in normalizer.py.

hgvs/normalizer.py
icebert: Yes. 3' shifting could result in insertions outside of the end of reference sequence, for example 'NM_212556.2:c.1400_1401insAC'.

I just updated this pull request to handle this issue.

PR #436, opened 2017-05-11
Migrate test runner to use pytest

lucaswiman:
Implements #343 (as a step towards #190) by adding a tox.ini file and migrating the test runner from nose to pytest. As an added benefit, `make test` now runs without manually setting up a virtualenv, as long as you have tox installed, which might allow simplifying the contributing file.

## Changes
- Migrate nose `attr` test tagging to use `pytest.mark`.
- Add tox.ini file with appropriate setup for running the tests on python 2.7.
- Update the `test` and `test-TAG` make targets
- Remove all remaining references to nose.

## Testing

I verified that `make test` and `make test-quick` both pass and run the expected tests.

PR #448, opened 2017-06-04
Add Python 3 Support

### LINKED ISSUE #190: add Python 3 support
Originally reported by: **Reece Hart (Bitbucket: [reece](http://bitbucket.org/reece), GitHub: [reece](http://github.com/reece))**

----------------------------------------

----------------------------------------
- Bitbucket: https://bitbucket.org/biocommons/hgvs/issue/190

lucaswiman:
Fixes #190. Refs #436 and #438. Most of the changes were fairly straightforward. The only tricky part was that the postgres rowdict object changed its layout between Python 2 and Python 3, necessitating a separate UTA cache file for each version. I'll add notes for the remaining changes worthy of comment inline.

I verified that `tox -e py36` passes. I haven't tried other versions of Python 3, though it would probably be worth supporting Python 3.5, and possibly 3.4. (3.3 is a bit of a stretch, and 3.2 doesn't even work with pip anymore.)


hgvs/config.py
```diff
@@ -65,7 +70,10 @@ def __getattr__(self, k):
         # Work around PyCharm bug https://youtrack.jetbrains.com/issue/PY-4213
         if k == "_cp":
             return
-        return ConfigGroup(self._cp[k])
+        try:
+            return ConfigGroup(self._cp[k])
+        except KeyError:
+            raise AttributeError(k)
```
lucaswiman: This broke where the test collector was calling `hasattr` on a config object. In python 2, hasattr returns False if _any_ exception is raised, while python 3 only catches `AttributeError`.

hgvs/decorators/lru_cache.py
```diff
@@ -32,6 +32,15 @@ def __init__(self, tup, hash=hash):
     def __hash__(self):
         return self.hashvalue
 
+    def __getstate__(self):
+        return True  # needs to be truthy for __setstate__ to be called
```
lucaswiman: This `__getstate__` and `__setstate__` setup prevents the `hashvalue` attribute from being pickled, which can lead to issues when hash randomization is enabled. The test suite is still unreliable with hash randomization enabled (not sure why), but this allowed more tests to pass. I think this is _technically_ unnecessary for this PR, but is an improvement I ended up implementing while trying & failing to make a py2/py3 compatible `cache.hdp` file.

hgvs/normalizer.py
```diff
@@ -215,15 +215,19 @@ def _get_boundary(self, var):
                 left = exon_starts[i]
                 right = exon_ends[i]
 
-                if var.posedit.pos.end.base - 1 < cds_start:
+                if cds_start is None:
```
lucaswiman: Python 3 doesn't consider `None` comparable with `int`, while Python 2 considers `None < i` for all ints `i`. This preserves the Python 2 behavior in this method, though I didn't evaluate the old behavior for correctness.

tests/test_clinvar.py
```diff
@@ -44,9 +44,14 @@ def test_clinvar(self, fn=data_fn, mod=None):
...
-        fh = LineIterator(fh=gzip.open(fn) if fn.endswith(".gz") else io.open(fn),
-                          skip_comments=True)
-        for rec in csv.DictReader(fh, delimiter=b"\t"):
+        if sys.version_info < (3, ):
+            fh = LineIterator(fh=gzip.open(fn) if fn.endswith(".gz") else io.open(fn),
+                              skip_comments=True)
+        else:
+            fh = LineIterator(fh=(io.TextIOWrapper(gzip.open(fn), encoding='utf-8')
+                                  if fn.endswith(".gz") else io.open(fn)),
+                              skip_comments=True)
```
lucaswiman: I didn't run the full `Test_Clinvar` tests because I don't have the local setup to run them quickly. I did run them for a while with no failures.

tests/test_hgvs_grammar_full.py
```diff
@@ -79,11 +83,11 @@ def test_parser_grammar(self):
                 is_valid = True if row["Valid"].lower() == "true" else False
 
                 for key in expected_map:
-                    expected_result = six.text_type(expected_map[key])
+                    expected_result = six.text_type(expected_map[key]).replace("u'", "'")
```
lucaswiman: The expected results included some tuples with `u"strings"` in them. This is a bit of a hack, though so was the previous approach. ¯\\\_(ツ)\_/¯

reece:
@lucaswiman Thanks for your Py3 efforts! I've been on (am still on) vacation. Reviewing now...

hgvs/normalizer.py
reece: This is almost certainly a thinko (and maybe a bug). Good catch.

lucaswiman:
Thanks @reece; I'm excited for when this gets released! Please feel free to ping me if you need any additional help on the Python 3 upgrade.

> I've been on (am still on) vacation.

😬 Sorry I interrupted your vacation. I certainly didn't mean to imply there was any rush on reviewing this. Hope you have a relaxing remainder of break.

PR #451, opened 2017-06-28
Add python 3 trove classifier.

lucaswiman:
@reece This PR includes two small changes I missed in #448:
- The trove classifiers for the package only included Python 2, which meant that tools like `caniusepython3` would return an incorrect status for `hgvs`.
- ~~Add Python 3 support to the changelog.~~


reece:
@lucaswiman: Would you please remove the ba41e97? The rst files are generated.

lucaswiman:
> The rst files are generated.

Ah OK. Done.

PR #457, opened 2017-08-30
Fix padding of nucleotide sequences in variant mapper

ianfab:
The padding in its current implementation throws an exception if it is executed, because `extend` always returns `None`, so `"".join(None)` fails with a `TypeError`.

### DIFF
--- a/hgvs/variantmapper.py
+++ b/hgvs/variantmapper.py
@@ -316,7 +316,7 @@ class VariantMapper(object):
                 # padding list so biopython won't complain during the conversion
                 tx_seq_to_translate = tx_seq[cds_start - 1:cds_stop]
                 if len(tx_seq_to_translate) % 3 != 0:
-                    "".join(list(tx_seq_to_translate).extend(["N"] * ((3 - len(tx_seq_to_translate) % 3) % 3)))
+                    tx_seq_to_translate += "N" * (-len(tx_seq_to_translate) % 3)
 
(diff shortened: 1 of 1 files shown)


reece:
Fabian-
Thanks for the patch.
Do you have an example of the issue? I don't think I see it and I'd like to
add a test case.
Thanks!
-Reece

ianfab:
@reece 
I encountered the issue when subclassing the data provider interface, but when using UTA I could not easily find an example where `len(tx_seq_to_translate) % 3 != 0` is satisfied, so the code line that causes the issue seems to be called only rarely.

To find examples, I used the following query for UTA:
```
select tx_ac, alt_ac, cds_start_i, cds_end_i, (cds_end_i - cds_start_i) % 3
from tx_def_summary_v
where (cds_end_i - cds_start_i) % 3 != 0
limit 10;
```

With the help of these results I was able to reproduce the `TypeError` with, e.g.:
```
import hgvs
import hgvs.parser
import hgvs.dataproviders.uta
import hgvs.assemblymapper

hgvsparser = hgvs.parser.Parser()
hdp = hgvs.dataproviders.uta.connect()
am = hgvs.assemblymapper.AssemblyMapper(hdp)

var_c = hgvsparser.parse_hgvs_variant('NM_002537.3:c.250del')
var_p = am.c_to_p(var_c)
```

reece:
This was merged into the 460-fix-padding branch. See #460 for comments about why I haven't merged this into master yet.

PR #467, opened 2017-10-05
466-Fix wrong mapping of identity variant

### LINKED ISSUE #466: Wrong variant mapping result for identity variant
For identity variant, the replace reference function in variant mapper only changes the ref of edit, causing the ref and alt discordant.

Example code:
```python
import hgvs.parser
import hgvs.assemblymapper
import hgvs.dataproviders.uta

hp = hgvs.parser.Parser()

hdp = hgvs.dataproviders.uta.connect()

am = hgvs.assemblymapper.AssemblyMapper(hdp, normalize=False)


var = 'NM_001292004.1:c.376='

var_c = hp.parse_hgvs_variant(var)

var_g = am.c_to_g(var_c)

str(var_g)
``` 
`NC_000005.10:g.74715659delins`

The correct mapping result should be `NC_000005.10:g.74715659=` instead of `NC_000005.10:g.74715659delins`.

icebert:
Fix #466


tests/test_issues_03xx.py
```diff
@@ -170,9 +170,9 @@ def test_379_move_replace_reference_to_variantmapper(self):
         # replace_reference code was in am, not vm. That meant that using vm directly
         # resulted in variants that were not reference corrected.
         g_var = self.hp.parse_hgvs_variant("NC_000006.11:g.44275011T=")
-        c_var = self.hp.parse_hgvs_variant("NM_020745.3:c.1015G>A")    # correct projection with ref replacement
```
icebert: I think the test case is wrong. The mapping result of identity variant should also be identity variant.

PR #516, opened 2018-08-25
hgvs.easy: Provide single-file import with usable defaults

reece:
@icebert : Please review this PR. See the top of `hgvs/easy.py` for description and examples.

I wasn't sure how hard it would be to implement lazy loading. It turns out that it was pretty easy (see hgvs/utils/lazywrapper.py). I should release that on its own!


icebert:
This PR looks good to me. I think it could be merged.

PR #527, opened 2018-10-04
Removing Biopython dependency

afrubin:
HGVS requires Biopython as a dependency, but the only things it uses are sequence translation and FASTA parsing. Since Biopython is so large and complex with lots of compiled code, it seemed quite heavyweight to require it for such simple functionality.

This PR and these two PRs to HGVS dependencies also in biocommons (https://github.com/biocommons/biocommons.seqrepo/pull/36 and https://github.com/biocommons/bioutils/pull/14) remove the Biopython requirement.


reece:
P.S. This was stalled until I got the new bioutils package on pypi. That's now available as bioutils 0.4.0, which means that we can refer to it in hgvs dependencies. So, the plan is to require bioutils > 0.4.0 and then release hgvs as a new minor version (not patch level). Sorry for the delay, and for not explaining the logistics.

afrubin:
No problem, I figured it was something like this. Thanks for the explanation!

reece:
Thanks for persisting, @afrubin . I'll watch the ci testing and merge when travis is green. (I'm not worried about the historical issues reported by DeepSource right now.)

afrubin:
Just wanted to follow up on this @reece. Travis is passing now, although I'm not sure what caused the coverage drop.

reece:
@afrubin: I apologize again for my tardiness in merging. I was on vacation when this came in and am just clearing out the backlog. Thanks for the PR, Alan!

PR #528, opened 2018-10-19
Installation Update

aaronslaff:
Look forward to hearing from you guys about what you think.


doc/installation.rst
```diff
@@ -1,116 +1,121 @@
...
 
 Supported Platforms
 @@@@@@@@@@@@@@@@@@@
 
-`hgvs` is developed primarily on Ubuntu systems and has been reported
-to work on Mac.  Other platforms and dependency versions are expected
-to work but have not been tested. Reports of successful operation on
-other platforms (and patches to enable this) are appreciated.
-**Python 2 is required.** (Python 3 support is in progress.)
+**Operating Systems: Ubuntu, MacOS**
```
reece: This section isn't quite right. hgvs works on Python 2.7 and 3.5+ now, and on Ubuntu and MacOS. It doesn't work on Windows (yet). On March 31, 2019, development will switch to Python 3.6+ only; no new features or bug fixes for Python 2.7-compatible versions of hgvs will be provided.

reece:
Thanks @aaronslaff. Overall, looks good.

Comments added -- see the Files Changed tab. 

Also, the build is now failing, which you can see above. Anything in the documentation that looks like code is tested so that the documentation doesn't have broken examples. I think your seemingly minor changes to the hgvs-shell session may be causing those tests to fail. You can edit, commit, and push again, then check back here for failures. Or, you should be able to run `make tox` locally. (Tip: run that now, before fixing, to ensure that you can reproduce the error)

doc/installation.rst
```diff
@@ -95,8 +97,8 @@ Fetch the source code::
 
 .. _seqrepo_install:
 
-Installing SeqRepo (recommended)
-@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+Installing SeqRepo (recommended, **required for Windows**)
```
reece: @aaronslaff: Other way around: seqrepo doesn't work on windows (at all), except in a docker image.

reece:
Please put these notebooks `examples/` with the others. Also, the filename convention there is lower case with hyphen as space. Please use that. (You'll find I'm not particularly dogmatic, but I do like consistency.)

reece:
This PR was abandoned and isn't easily merged now. Discarding. (Sorry @aaronslaff )

Issue #570, opened 2019-09-25
PG schema search path is set more often than necessary


csw:
The PostgreSQL schema search path is set every time a new cursor is acquired, which means executing an extra DB statement every time. This only needs to happen once, when we get a new connection.

PR #571, opened 2019-09-25
Set search path only when necessary

csw:
See #570; the PostgreSQL schema search path only needs to be set the first time we use a connection, not every time we acquire a cursor.


reece:
Apologies for the long wait. CI tests were failing, and I wanted that resolved before merging.

I've figured out what broke in tests and manually merged this change into master.

reece:
Thanks for the PR, @csw .

PR #572, opened 2019-09-26
Speed up tx_for_region query using materialized view

csw:
The tx_for_region query used by AssemblyMapper.relevant_transcripts can be a bottleneck in some cases; rewriting it to use a materialized view already in the UTA schema can give about a 4x speedup, and 16x with an additional index.

This query is PostgreSQL-specific, so this leaves the current one in place for other databases and puts the new query in a separate copy of `_queries` in UTA_postgresql.

Adding a functional index is ideal for this:

```sql
create index exon_set_exons_fp_mv_ac_method_pos_ix
on exon_set_exons_fp_mv (
  alt_ac,
  alt_aln_method,
  (starts_i[case when alt_strand=1 then 1 else array_length(starts_i, 1) end]),
  (ends_i[case when alt_strand=1 then array_length(ends_i, 1) else 1 end])
);
```

### DIFF
--- a/hgvs/dataproviders/uta.py
+++ b/hgvs/dataproviders/uta.py
@@ -147,6 +147,8 @@ class UTABase(Interface):
             from transcript T
             join exon_set ES on T.ac=ES.tx_ac where alt_aln_method != 'transcript' and hgnc=?
             """,
+        # ensure this has the same semantics as the PostgreSQL-specific
+        # tx_for_region query in UTA_postgresql
         "tx_for_region":
         """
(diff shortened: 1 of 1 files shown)


reece:
@csw : Thanks for these PRs! I see that travis is choking, apparently for reasons unrelated to the PRs.  I can't tell whether these are building. I'm a bit jammed up for time right now, but will endeavor to get to these in the next week. Again, thanks.

reece:
Thanks @csw. 

This PR is really better targeted for UTA.

Although I'm not merging this, I have made note of the use case in notes for UTA overhaul.

PR #575, opened 2019-11-22
Update setup.py ipython version requirement

jamespeacock:
Remove python2 compatibility version enforcement on ipython. 
Python2 support will be dropped in 1 month and ipython should not be locked at <6.0


jamespeacock:
@reece Hoping this can be a quick PR. Let me know your thoughts. If you'd like me to work on fully dropping python2 support in this library in the next month I can do so as well.

reece:
Thanks for the motivation to kick 2.7 to the curb.

Will merge despite travis failures, which it turns out appear to also be related to Py 2.7 legacy.

PR #601, opened 2020-06-03
fix formatting of example docker command

alexhenrie:
Without the second colon, the two dashes in the argument `--name` are combined into a single long dash (i.e. `–name`), which causes a syntax error when copying and pasting the command.


reece:
Good catch. Thanks!

alexhenrie:
Thank you!

Issue #606, opened 2020-07-30
Index Error for variant using am37.c_to_g()


sptaylor:
We encountered a bug when running c_to_g via assembly mapper for variant 'NM_001291722.1:c.283-3C>T':
```
$ hgvs-shell
Python 3.6.9 (default, Jan  8 2020, 08:00:01)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.


############################################################################
hgvs-shell -- interactive hgvs
hgvs version: 1.5.1
data provider url: postgresql://anonymous:anonymous@uta.biocommons.org/uta/uta_20180821
schema_version: 1.1
data_version: uta_20180821
sequences source: SeqRepo (/Users/paige.taylor/Data/seqrepo/2018-08-21)

The following variables are defined:
* global_config
* hp, parser, hgvs_parser -- Parser instance
* hdp, hgvs_data_provider -- UTA data provider instance
* vm, variant_mapper, hgvs_variant_mapper -- VariantMapper instance
* am37, hgvs_assembly_mapper_37 -- GRCh37 Assembly Mapper instance
* am38, projector, hgvs_assembly_mapper_38 -- GRCh38 Assembly Mapper instances
* hn, normalizer, hgvs_normalizer -- Normalizer instance
* hv, validator, hgvs_validator) -- Validator instance

The following functions are available:
  * parse, normalize, validate
  * g_to_c, g_to_n, g_to_t,
  * c_to_g, c_to_n, c_to_p,
  * n_to_c, n_to_g,
  * t_to_g,
  * get_relevant_transcripts

When submitting bug reports, include the version header shown above
and use these variables/variable names whenever possible.



In [1]: var = hp.parse_hgvs_variant('NM_001291722.1:c.283-3C>T')

In [2]: am37.c_to_g(var)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
~/virtualenv/hgvs_v15/lib/python3.6/site-packages/hgvs/shell.py in <module>
----> 1 am37.c_to_g(var)

~/virtualenv/hgvs_v15/lib/python3.6/site-packages/hgvs/assemblymapper.py in c_to_g(self, var_c)
    114         alt_ac = self._alt_ac_for_tx_ac(var_c.ac)
    115         var_out = super(AssemblyMapper, self).c_to_g(
--> 116             var_c, alt_ac, alt_aln_method=self.alt_aln_method)
[...]

PR #609, opened 2020-12-17
issue #606 - catch index error for variants at alignment gap

kyuhas:
Hi Reece, 

I tried adding unit tests in test_hgvs_variantmapper.py for the scenarios described in issue 606. However, I couldn't add them because of the cache that was being used. I was unable to open the cache-py3.hdp file. Would you be able to help me with this? I'd like to add a couple of tests to confirm that the correct exception is being thrown.

Thanks!
Kaylee

### DIFF
--- a/hgvs/variantmapper.py
+++ b/hgvs/variantmapper.py
@@ -21,7 +21,7 @@ import hgvs.utils.altseqbuilder as altseqbuilder
 import hgvs.sequencevariant
 import hgvs.validator
 
-from hgvs.exceptions import HGVSDataNotAvailableError, HGVSUnsupportedOperationError, HGVSInvalidVariantError, \
+from hgvs.exceptions import HGVSUnsupportedOperationError, HGVSInvalidVariantError, \
     HGVSInvalidIntervalError
 from hgvs.decorators.lru_cache import lru_cache
 from hgvs.enums import PrevalidationLevel
@@ -162,7 +162,8 @@ class VariantMapper(object):
                                              alt_ac=var_g.ac,
                                              alt_aln_method=alt_aln_method)
 
-        if (mapper.strand == -1 and not hgvs.global_config.mapping.strict_bounds
+        if (mapper.strand == -1
+                and not hgvs.global_config.mapping.strict_bounds
                 and not mapper.g_interval_is_inbounds(var_g.posedit.pos)):
             _logger.info("Renormalizing out-of-bounds minus strand variant on genomic sequence")
             var_g = self.left_normalizer.normalize(var_g)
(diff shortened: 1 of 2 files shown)


reece:
Happy to help! Thanks for give this a shot.

Did you find/confirm the underlying cause? Do you have test cases? Please see tests/issues/test_* files for examples and put your tests in a new file. Also, a diagram of the issue would be very valuable. (test_437.py has a good example)

You don't edit cache-py3.hdp directly. Instead, you run tests in a special mode to add new data required for tests. For example: 

```
$ HGVS_CACHE_MODE=learn make test
$ HGVS_CACHE_MODE=learn pytest tests/issues/test_606.py
```

The learn mode will fetch and cache and new data.

kyuhas:
Thanks Reece, that was super helpful! While I don't think this code change solves the core issue (which seems to be related to the _get_altered_sequence method), this at least solves my use case of throwing a more meaningful error. So this wouldn't close issue 606 (since the core issue isn't addressed), but it would make the integration with my code a little easier. I'll be working on some diagrams and will update this PR early next week. 

Thanks!

reece:
Hi @kyuhas. I would prefer to understand the root cause and solve that once rather than expect all callers to wrap. I look forward to seeing your assessment of what's happening and appreciate your time.

kyuhas:
Hi Reece,
I have updated the pull request to (1) add unit tests, and (2) move the try/catch for Index Errors into the "_get_altered_sequence" method where the error occurs. _get_altered_sequence is only used in variantmapper.py, and only when the variant is at an "alignment gap".  So the idea is that if we already know something is funky with the alignment, and then you end up with an index error, throwing the HGVSInvalidIntervalError seems appropriate.  

There seems to be multiple root causes possible, two of which are illustrated by the unit tests.  

The first unit test example, NC_000001.10:g.16890441C>G has transcript NM_017940.4 with 25 exons in the 'transcript' alignment to itself, but 28 exons in the 'splign' alignment to the reference sequence. So in this case, ultimately the root cause is inconsistent data. 

The second unit test example, NM_001291722.1:c.283-3C>T, seems to be invalid input (this is the example from issue 606 https://github.com/biocommons/hgvs/issues/606).  According to the transcript in UTA, exon 5 starts at position 286 of the transcript, not position 283.  The input "NM_001291722.1:c.286-3C>T" does not give an index error. 

In both cases, in the "_get_altered_sequence" method, either bad data or bad input ultimately causes pos_start/pos_end to be outside of the bounds of the seq array, causing the index error.

I am not yet familiar enough with hgvs to identify if there is a more appropriate spot to identify the bad data/bad input-- it will take a while to fully understand.

Unfortunately, we have a time-sensitive need to handle this error.  Let me know if this updated pull request is a good fix for now. If not, then I'll have my code catch Index Errors when calling the hgvs package until this is resolved some time in the future. 

Thanks!

reece:
This issue was closed by stalebot. It has been reopened to give more time for community review.  See [biocommons coding guidelines](https://biocommons.org/contributing/coding-guidelines/) for stale issue and pull request policies.  This resurrection is expected to be a one-time event.

Issue #628, opened 2021-10-29
Add support for pulling from seqrepo-rest-service.


andreasprlic:
**Summary**
It would be nice to be able to connect hgvs to pull sequence data from an instance of [seqrepo-rest-service](https://github.com/biocommons/seqrepo-rest-service), rather than a local copy of the data (or NCBI). Happy to provide a PR for that. 

**Implementation Sketch:**

Make the [vrs-python data proxy](https://github.com/ga4gh/vrs-python/blob/0c80d489004d9abdbc9e180568bbf92e21f9c2b2/src/ga4gh/vrs/dataproxy.py) more re-usable. Where is the best location for this? Should this be directly in seqrepo? Alternatively, this could also reside in [bioutils](https://github.com/biocommons/bioutils). Any recommendations? (otherwise I'll just create a PR in seqrepo for this), and then a PR in hgvs to update the [SeqFetcher](https://github.com/biocommons/hgvs/blob/15c393741ee5064e64fd4df0ac687d69a315520a/hgvs/dataproviders/seqfetcher.py) to use this new proxy as a 3rd mechanism for sequence access.

reece:
@andreasprlic Awesome that you're working on this. It's been needed for a long time. (See #199)

Architectural question: Would it be better (and do we have time) to implement something more like seqrepo-rest-service as a biocommons-rest-service, which would encapsulate UTA and seqrepo? hgvs would then need a new biocommons-rest data provider. And, we'd have to replumb hgvs to standardize responses from both sources. I think this route is better, but also larger, which is why I haven't done it. :-/

andreasprlic:
@reece yes I totally agree. It would be better to have a service layer also on top of UTA.  Having said that, I don't think I have time right now to work on that. Perhaps we should write this up as a ticket (or an epic).

In the meanwhile I was hoping that https://github.com/biocommons/biocommons.seqrepo/pull/98 could be a first step in the right direction.

bmrobin:
> Would it be better (and do we have time) to implement something more like...

@reece - andreas created this github issue after a conversation he & i had about how a requirement has arisen at Invitae with a project i'm working on that requires this now, and ultimately may be of use architecturally for future projects at Invitae.

PR #632, opened 2021-12-13
#628: add support for SeqFetcher to use seqrepo-rest-service as data source

bmrobin:
https://github.com/biocommons/hgvs/issues/628


bmrobin:
@andreasprlic fyi as well

andreasprlic:
does this also need a seqrepo library upgrade?

bmrobin:
> does this also need a seqrepo library upgrade?

pushed 👍

andreasprlic (approved):
Thanks, LGTM!

PR #643, opened 2022-11-07
Improve uta pooling

jdasilva-invitae:
We've recently had some issues while using hgvs. Sometimes when UTA receives an update (or it experiences some issue) it closes all connections and this causes some of our applications to experience `psycopg2.pool.PoolError`. 

Looking into it, we realized this portion of the code closes the connections when `self.pooling` and then it tries to connect again, this is essentially what is causing the PoolError.

So if `self.pooling` we just update the pool connections (as the snippet above does) and if not we reconnect to the UTA database. 

We've also tested this locally with our applications and the error stopped showing up. 

To be more specific, here's the source code for psycopg2 https://github.com/psycopg/psycopg2/blob/master/lib/pool.py

The first thing it does when either getting a conn or updating the conn is to check whether the pool is closed (which we are currently doing in `psycopg2.OperationalError` in our `_get_cursor` method

```
if self.closed:
            raise PoolError("connection pool is closed")
```

### DIFF
--- a/hgvs/dataproviders/uta.py
+++ b/hgvs/dataproviders/uta.py
@@ -623,8 +623,9 @@ class UTA_postgresql(UTABase):
                 _logger.warning(
                     "Lost connection to {url}; attempting reconnect".format(url=self.url))
                 if self.pooling:
-                    self._pool.closeall()
-                self._connect()
+                    self._pool.putconn(conn)
+                else:
+                    self._connect()
                 _logger.warning("Reconnected to {url}".format(url=self.url))
 
             n_tries_rem -= 1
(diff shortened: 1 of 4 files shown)


hgvs/dataproviders/uta.py
```diff
@@ -623,8 +623,9 @@ def _get_cursor(self, n_retries=1):
                 _logger.warning(
                     "Lost connection to {url}; attempting reconnect".format(url=self.url))
                 if self.pooling:
-                    self._pool.closeall()
-                self._connect()
+                    self._pool.putconn(conn)
+                else:
+                    self._connect()
                 _logger.warning("Reconnected to {url}".format(url=self.url))
```
cpeel, line 629: I would suggest indenting this so it only outputs the warning in the non-pooled case after the `_connect()`. In the pooled case we aren't actually reconnected here, we're just prepared for a reconnect on the next use.

andreasprlic (approved):
Thanks, LGTM

cpeel (approved):
Thank you for your work on this!

PR #646, opened 2022-11-12
630 migrate to GitHub actions for testing and packaging

reece:
This issue turned out to be much larger than anticipated:

- hgvs hadn't been migrated to github actions, so there was no immediate route to testing and packaging
- While working on actions, I found deprecation errors in the existing packaging tooling, so I updated the packaging tooling to use a more modern setup (pyproject.toml, setup.cfg) 
- Once that was done, I found that tests didn't work due to various debt (like uta_dev schema isn't publicly accessible any longer)
- In the course of fixing that, I discovered that the data cache had a bug that led to lots of run-time misses (and slowness)
- And while fixing that, I saw that tests initialized connections many more times than necessary, which also made testing slow

Whew!

So, this PR does way more than any single PR should. It:

- Adds a github action
- Significantly updates the packaging tooling
- Reduces the number of hdp initialization calls, leading to speedups
- Rebuilds the test cache

Using this branch, I tagged 1.5.3.rc2, which successfully ended up in https://pypi.org/project/hgvs/1.5.3rc2/. :-)


pyproject.toml
```diff
@@ -0,0 +1,109 @@
+[project]
+name = "hgvs"
+authors = [
+  { name="HGVS Contributors", email="biocommons-dev@googlegroups.com" },
+]
+description = "HGVS Parser, Formatter, Mapper, Validator"
+readme = "README.rst"
+license = { file="LICENSE.txt" }
+requires-python = ">=3.8"
```
jdasilva-invitae, line 9: You might want to update the pypi docs
```
Important: biocommons packages require Python 3.6+. [More](https://groups.google.com/forum/#!topic/hgvs-discuss/iLUzjzoD-28)
```

Was there a similar discussion around bumping this to 3.8? We are using hgvs in some projects that are on 3.6

cc @reece @andreasprlic

jdasilva-invitae: Any news on it @reece ?

athaevar: Without bumping the major version, this change can be breaking for anyone using hgvs. @reece would you mind reverting to 3.6? We dont see any apparent code that requires 3.8 features.

PR #647, opened 2022-11-16
Go back to py36

jdasilva-invitae:
Moving from 1.5.2 to 1.5.3 python was bumped from 3.6 to [3.8+](https://github.com/biocommons/hgvs/pull/646/files#:~:text=requires%2Dpython%20%3D%20%22%3E%3D3.8%22), was this a necessary change? We have a lot of repo's that use hgvs and are still on 3.6.x

cc @reece 

---

Also fixed the Makefile for pytest, since now we don't run `python setup.py pytest`, and raw `pytest` doesn't have the `--adopts` arg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,13 +75,13 @@ build: %:
 test:
 	pytest
 test-code:
-	pytest --addopts="${TEST_DIRS}"
+	pytest ${TEST_DIRS}
 test-docs:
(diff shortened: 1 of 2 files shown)


bpeterman (approved):

andreasprlic (approved):
Thanks, LGTM

Issue #653, opened 2023-05-22
Babelfish.hgvs_to_vcf tests against length_change signature not method result


davmlaw:
```hgvs.posedit.PosEdit.length_change``` is a method, while ```hgvs.extras.babelfish.Babelfish.hgvs_to_vcf```

performs a test against the signature:

```
if end_i - start_i == 1 and vleft.posedit.length_change == 0:
```

So this always fails. This makes a conversion wrong, eg one output is:

```
assert _h2v("NC_000006.12:g.49949407A>T") == ('6', 49949406, 'AA', 'AT', 'sub')
```

Fix is to basically add a method call, ie:

```
if end_i - start_i == 1 and vleft.posedit.length_change() == 0:
```

It's also hardcoded for GRCh38, ie doesn't use the passed assembly_name.

I'll make a pull request for these fixes shortly

davmlaw:
Hi - I'm going to close this, and implement a pull request under #447

Issue #447, opened 2017-05-29
Implement hgvs babelfish: hgvs ↔ {vcf, vmc, spdi, beacon, brcaexchange}


reece:
There are several alternative variant formats. This issue intends to implement a module to facilitate conversions between those and the hgvs data model.

Planned formats:
* vcf
* [vmc](https://github.com/ga4gh/vmc)
* spdi
* beacon
* brcaexchange

Considerations:
* How to support assemblies? I'm loathe to emit strings like `chr1:1234:A>T` because they're underspecified.
* VCF should be 5'-shuffled

reece:
See hgvs.extras.babelfish.py (created for #243).

lazobax:
For people finding this who are unfamiliar with the codebase or are beginner coders, let me save you some time.

Huge thanks for @davmlaw and the rest of the team for finally implementing the hgvs to vcf conversion, god knows it was painful to do it manually in 2021. Glad to have found this now. 

If you want to get straight to converting, do this: 

1.  copy the `hgvs/src/hgvs/extras/babelfish.py` file locally. 
2. add the following to it:

``` 
if __name__ == "__main__":
    import sys
    import hgvs.dataproviders.uta 
    from hgvs.exceptions import HGVSError

    assembly_name = sys.argv[1]
    hgvs_str = sys.argv[2]
    hp = hgvs.parser.Parser()
    var_g = hp.parse_hgvs_variant(hgvs_str)
    hdp = hgvs.dataproviders.uta.connect()
    bf = Babelfish(hdp, assembly_name)
    vals = bf.hgvs_to_vcf(var_g)
    print(vals)
```
3. call the script like this `python3 babelfish.py <assembly_name> <HGVS_var>` and it should print the vcf fields you need to the terminal. A list of valid assembly names can be found at `hgvs/lib/python3.12/site-packages/bioutils/_data/assemblies` locally. 

This is a sloppy attempt at getting it to work fast which is what a lot of you will be looking for if you end up here. However, it's better to just instantiate it as an object in your main script. 

I will also be testing this further to see if it's robust. We need more cross-format compatibility in the field.

PR #654, opened 2023-05-22
#653 - Babelfish.hgvs_to_vcf tests against length_change signature no…

davmlaw:
Fix for #653

### DIFF
--- a/src/hgvs/extras/babelfish.py
+++ b/src/hgvs/extras/babelfish.py
@@ -1,6 +1,6 @@
 """translate between HGVS and other formats"""
 
-import bioutils.assemblies
+from bioutils.assemblies import make_ac_name_map
 
 import hgvs.normalizer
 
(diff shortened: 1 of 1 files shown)


reece (approved):
Sorry for the delay, @davmlaw. I don't know how I missed this PR

davmlaw:
I've been working in this area, doing full babelfish VCF conversion implementation - so it might be best to wait for me to finish that, as I ended up deleting this line of code.

I think left anchor (same staring base) should be decided by ```typ in ('del', 'ins')``` rather than the length of the change being non-zero, because a delins can have a non-zero change but doesn't have the same starting base, eg:

NC_000003.12:g.128483940_128483945delinsC

The ```end_i - start_i == 1 and vleft.posedit.length_change() == 0``` code would convert it to: 

3:128202782 TGGCCGG>TC

While ```typ in ('del', 'ins')``` would put it as:

3:128202783 GGCCGG>C

If you run the first one through [VT Normalization](https://genome.sph.umich.edu/wiki/Variant_Normalization) it converts it to the second

davmlaw:
I'll close this and make a new pull request under [#447](https://github.com/biocommons/hgvs/issues/447)

PR #657, opened 2023-05-31
Added a few REST methods

ccaitlingo:


davmlaw:
It looks like you've implemented a server.

The HGVS package would be a client for a REST server, ie code to contact the server and consume the response

The server should be in a completely separate project - remember the goal is to reduce the number of dependencies

[Here is my JSON client data provider for example](https://github.com/SACGF/cdot/blob/main/cdot/hgvs/dataproviders/json_data_provider.py#L371)

I think you should think about the structure of JSON before writing APIs, rather than just translating methods into API calls, because you can combine some (reducing a round trip across the network is a big saving)

ccaitlingo:
@davmlaw Yes, Reece pointed out too that this should be done as a separate project.  Thanks-- I've made a new project now.  In terms of JSON structure, I'm thinking of using BaseModel to format the responses.  Does this align with what you're envisioning?  I know you talked about wanting to make a common format like [this from cdot](http://cdot.cc/transcript/NM_001001890.3).  

Also, that is a really good point you make about combining trips!  I'll spend some time looking at code on the HGVS side of things to see what type of requests would be convenient to combine into one response.  Thanks! :)

reece:
@ccaitlingo Okay to close this PR?

ccaitlingo:
Yep!

PR #659, opened 2023-06-12
Added deprecation notice

ccaitlingo:


reece (approved):
looks good to me

reece:
Before integrating this notice, we need to provide an affirmative alternative to clients. I'll close this PR now, but expect to revisit a similar idea eventually.

PR #660, opened 2023-06-13
#447 - implement babelfish for VCF

### LINKED ISSUE #616: Babelfish using "GRCh38" when it should be using "assembly_name"
I think the last line in  

```class Babelfish:
    def __init__(self, hdp, assembly_name):
        self.hdp = hdp
        self.hn = hgvs.normalizer.Normalizer(hdp,
                                             cross_boundaries=False,
                                             shuffle_direction=5,
                                             validate=False)
        self.ac_to_chr_name_map = {
            sr["refseq_ac"]: sr["name"]
            for sr in bioutils.assemblies.get_assembly("GRCh38")["sequences"]}
```

should be

`for sr in bioutils.assemblies.get_assembly(assembly_name)["sequences"]}`

davmlaw:
#447 - Implement hgvs babelfish for VCF

Also fixes

#616 - Babelfish hardcoded GRCh38

I changed the test expected VCF results - my reasoning follows:

NC_000006.12:g.49949407=
Old: None
New: ("6", 49949407, "A", ".", 'identity')
Notes: As we can now load VCFs, this is now symmetrical (ie can convert to and back again w/o change). You can represent an identity/no change this way in VCF or by repeating the ref, I went with '.' but feel free to change

The other changes involve VT parsimony rules - https://genome.sph.umich.edu/wiki/Variant_Normalization

"A variant has superfluous nucleotides on its left side if the leftmost nucleotide of each variant is of the same type and the removal of the nucleotide from each allele will not result in an empty allele."

NC_000006.12:g.49949407A>T
Old: ("6", 49949406, "AA", "AT", "sub")
New: ("6", 49949407, "A", "T", "sub")

NC_000006.12:g.49949413_49949414delinsCC
Old: ("6", 49949412, "AAA", "ACC", "delins")
New: ("6", 49949413, "AA", "CC", "delins")


src/hgvs/extras/babelfish.py
```diff
@@ -17,12 +25,17 @@ def _as_interbase(posedit):
 
 
 class Babelfish:
+    """ The plan here is to get this into biocommons HGVS via pull request
```
reece, line 28: Once merged and released, this comment no longer applies, right? If so, plea remove.

tests/test_hgvs_extras_babelfish.py
reece, line 1: I ❤️ tests. Thank you.

reece (commented):
Minor comment about a comment. Otherwise, looks great. Thanks!

src/hgvs/extras/babelfish.py
davmlaw: ha ha, whoops yeah I copy/pasted it into my code to test it for a while

davmlaw:
Hi @reece are you able to review?

Afterwards I'll start on the parser and do as a separate PR

reece (approved):
lgtm. Thanks @davmlaw . Sorry for the long delay.

PR #662, opened 2023-06-30
modernize: Python version support, CI/CD, & build

dmyersturnbull:
- [x] Test and document Python 3.11 support. Also drop past-EOL 3.6 and 3.7.
- [x] Fix small issues in docs
- [ ] Re-enable CI: I noticed the Travis tests no longer run. (I'm guessing when they changed to .com.) Would GitHub Actions be ok for this? I can add a new test workflow.
- [ ] pre-commit? I found that having tools like black run on commit can clean up diffs. If the devs like the idea, I'll add it.
- [ ] replace [https://github.com/readthedocs/sphinx_rtd_theme](rtd_theme)? It's now poorly maintained, and [furo](https://github.com/pradyunsg/furo) may be a good replacement.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,7 +14,7 @@ jobs:
   #     - name: Set up Python
   #       uses: actions/setup-python@v4
   #       with:
-  #         python-version: "3.10"
+  #         python-version: "3.11"
   #         cache: "pip"
   #         cache-dependency-path: "**/setup.cfg"
 
@@ -43,7 +43,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     steps:
       - uses: actions/checkout@v3
@@ -75,7 +75,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.11"
           cache: pip
           cache-dependency-path: "**/setup.cfg"
 
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,9 +7,9 @@ version: 2
 
 # Set the version of Python and other tools you might need
 build:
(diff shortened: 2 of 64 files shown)


korikuzma:
+1 for pre-commit. We use this in our repos and it's really nice.

jsstevenson:
I really like Furo as well -- though given that biocommons.org is using Material for Mkdocs, maybe it'd be better to standardize biocommons docs around that?

reece:
Hi @dmyersturnbull : Awesome. All of these are great directions. The only input I have is that I'd prefer to see these changes as independent PRs. There's simply no reason to bundle them and small PRs are much easier (and safer) to consider.

https://github.com/biocommons/biocommons.example is intended to be the upstream template for all repos to ensure consistency. A better approach to your laudable goals would be to add them, one-by-one, to the example repo so that we can distribute them to all of biocommons projects. (e.g., commit hooks, black, etc.).

Thanks!

reece (commented):
Hi @dmyersturnbull : Awesome. All of these are great directions. The only input I have is that I'd prefer to see these changes as independent PRs. There's simply no reason to bundle them and small PRs are much easier (and safer) to consider.

https://github.com/biocommons/biocommons.example is intended to be the upstream template for all repos to ensure consistency. A better approach to your laudable goals would be to add them, one-by-one, to the example repo so that we can distribute them to all of biocommons projects. (e.g., commit hooks, black, etc.).

Thanks!

reece:
> I really like Furo as well -- though given that biocommons.org is using Material for Mkdocs, maybe it'd be better to standardize biocommons docs around that?

Some docs are in sphinx. I've been using mkdocs for the boicommons.org web pages. IMO, sphinx stinks for web sites and mkdocs stinks for complex docs (like those in hgvs), so for the moment I'm okay with having both and using the better tool for the job.

reece:
This issue was closed by stalebot. It has been reopened to give more time for community review.  See [biocommons coding guidelines](https://biocommons.org/contributing/coding-guidelines/) for stale issue and pull request policies.  This resurrection is expected to be a one-time event.

PR #669, opened 2023-08-03
Ensure hgvs seqrepo dir is unset during testing

ccaitlingo:


davmlaw:
Looks like you should delete the setenv on the previous line? And there is a duplicate line?

davmlaw:
Ah sorry ignore my comment I was reading using my phone on the bus and thought the variable was the same ("_URL" and "_DIR" )

reece (approved):

PR #673, opened 2023-08-26
Add ruff linter configuration

ecalifornica:
Per @korikuzma's suggestion on #671, this adds a basic `ruff` configuration. This should be merged after the 18 errors are fixed.

Related:
#413

PR #676, opened 2023-08-26
Fix linter complaint about ambiguous variable name

ecalifornica:
Fixes linter error E741 raised by `ruff`. I don't know why `ruff` wasn't upset by the variable `o`. The variable name `line_length` is only a guess on my part.

Unblocks #673


reece (commented):
I'd like to leave this as-is.

reece (commented):
Please add a noqa tag instead

reece (approved):

PR #679, opened 2023-08-26
Fix undefined names

ecalifornica:
Fixes linter error F821 raised by `ruff`. The line `from hgvs.dataproviders.ncbi import connect` is only a guess on my part.

Unblocks #673

### DIFF
--- a/src/hgvs/dataproviders/uta.py
+++ b/src/hgvs/dataproviders/uta.py
@@ -103,9 +103,7 @@ def connect(db_url=None, pooling=hgvs.global_config.uta.pooling, application_nam
         _logger.warning(f"You are executing tests using remote data ({db_url})")
 
     url = _parse_url(db_url)
-    if url.scheme == "sqlite":
-        conn = UTA_sqlite(url, mode, cache)
-    elif url.scheme == "postgresql":
+    if url.scheme == "postgresql":
         conn = UTA_postgresql(url=url, pooling=pooling, application_name=application_name, mode=mode, cache=cache)
     else:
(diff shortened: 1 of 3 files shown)


src/hgvs/dataproviders/uta.py
```diff
@@ -624,6 +624,11 @@ def _set_search_path(self, cur):
         cur.execute("set search_path = {self.url.schema},public;".format(self=self))
 
 
+class UTA_sqlite(UTABase):
```
reece, line 627: UTA_sqlite was removed long ago. I think the origin of this concern is uta.py:107, with a case for a sqlite scheme. We should remove that case rather than add an empty class.

reece (changes requested):

tests/test_hgvs_alignmentmapper.py
ecalifornica, line 1: Added these changes since last reviewed, had neglected to run `ruff` on the `tests/` path.

reece (approved):

src/hgvs/normalizer.py
```diff
@@ -13,9 +13,11 @@
 import hgvs
 import hgvs.validator
 import hgvs.variantmapper
+from hgvs.dataproviders.ncbi import connect
 from hgvs.exceptions import (HGVSDataNotAvailableError,
                              HGVSInvalidVariantError,
                              HGVSUnsupportedOperationError)
+from hgvs.parser import Parser
```
davmlaw, line 20: This introduced a circular dependency from Parser <-> normalizer

Issue #670, opened 2023-08-09
HGVSParseError for gene symbols with "-" (dashes) in them


davmlaw:
In theory, gene names shouldn't have dashes in them, but in the wild they do, and "ADAMTSL4-AS1" is an official [HGNC:32041 symbol](https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:32041)

```
from hgvs.parser import Parser
parser = Parser()
parser.parse_hgvs_variant("NR_104133.1(ADAMTSL4-AS1):n.101del")
```

gives:

```
ParseError: 
NR_104133.1(ADAMTSL4-AS1):n.101del
                    ^
Parse error at line 1, column 20: expected one of ')', or a letter or digit. trail: [letterOrDigit]
```

while removing the dash, eg

```
parser.parse_hgvs_variant("NR_104133.1(ADAMTSL4):n.101del")
```

works fine

davmlaw:
Looks like we may need "_" (underscore) as well, here are some other approved HGNC symbols that currently fail:

[GTF2H2C_2](https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:44196)
[APOBEC3A_B](https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:44196)
[C4B_2](https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:42398)

davmlaw:
Much like the previous gene_symbol match disallows numbers at the start of a gene name, I wanted to disallow dashes/underscores at the start and end. I tried to fix this using the obvious regex style:

```
gene_symbol = <letter ('_' | '-' | letterOrDigit)* letterOrDigit+>
```

However I believe the Ometa parser has no backtracking on success (unlike regex which will attempt re-matches if it fails at a higher level).

So - instead I had to allow the dashes/underscore at the end, but then used Python code to validate it

I have done this in a branch off a branch of #661 (store generated Parser code)

https://github.com/davmlaw/hgvs/commit/c5bdd9bb379f22600705b6541bbba6384f083bbd

So I won't do this pull request until that one is merged.

If you reject that one, and want this one against biocommons HGVS master (ie no stored generated Parser Python code) then let me know

davmlaw:
Will make pull request this afternoon (6-8 hours from now)

davmlaw:
[...]

PR #683, opened 2023-09-11
#670 - Gene Symbol support dash and underscore

davmlaw:
For issue #670 - some HGNC symbols have dashes/underscores


reece:
I think this is fine. However, would you please also consider and comment on the following alternative?

```
>>> import parsley
>>> g = parsley.makeGrammar("gene_symbol = <letter (letterOrDigit | ('-'|'_') ~~letterOrDigit)+>", {})
>>> def parseable(g, s):
...     try:
...         g(s).gene_symbol()
...         return True
...     except parsley.ParseError:
...         return False
... 
>>> for s in "ABC123 ABC_123 ABC-123 ABC__123    _ABC123 ABC123_ ABC/123".split():
...     print(f"{s}: {parseable(g,s)}")
... 
ABC123: True
ABC_123: True
ABC-123: True
ABC__123: False
_ABC123: False
ABC123_: False
ABC/123: False
```

The essential newness here is that the -_ requires a positive lookahead for a letter or digit, so therefore can't end an expression

Whaddya think?

davmlaw:
Thanks @reece I had tried to do that but couldn't work out how.

Checked and tests are good, so have applied your change and updated pull request

reece:
@davmlaw The PR looks the same to me. It doesn't look like your grammar rule is updated and it still uses `validate_gene_symbol`. Did you push?

davmlaw:
@reece - sorry branching issue on my end, I think it's good now

davmlaw:
@reece - I've found an issue - it doesn't support matching a gene without a transcript if it has a dash - eg "SP2-DT:c.1G>T" fails

reece (approved):

Issue #689, opened 2023-09-13
Cannot import Parser due to circular import


davmlaw:
The [linting change](https://github.com/biocommons/hgvs/pull/679/commits/5a9002a881116951ebc18f79b327365e18bc3bab#r1323788197) introduced a circular dependency 

```
from hgvs.parser import Parser
```

Now dies with:

```
/usr/local/lib/python3.10/dist-packages/hgvs/normalizer.py in <module>
     19     HGVSUnsupportedOperationError,
     20 )
---> 21 from hgvs.parser import Parser
     22 from hgvs.utils.norm import normalize_alleles
     23 

ImportError: cannot import name 'Parser' from partially initialized module 'hgvs.parser' (most likely due to a circular import) (/usr/local/lib/python3.10/dist-packages/hgvs/parser.py)
```

This should be an easy fix, just import in the main section instead of the module

davmlaw:
tested - fixed

PR #690, opened 2023-09-13
#689 - undo circular import introduced from linting fix

davmlaw:
Fix for #689 - cannot import parser due to circular import introduced in #679 - "Fix undefined names"


src/hgvs/normalizer.py
```diff
@@ -416,6 +415,8 @@ def _normalize_alleles(self, var, boundary):
 
 
 if __name__ == "__main__":
+    from hgvs.parser import Parser
```
larrybabb, line 418: per @reece "we should remove this block and everything will work fine."

PR #693, opened 2023-09-17
local installation readme updates

katiestahl:
I am so sorry for the reformatting/whitespace changes :' ) but also I am a huge proponent of removing trailing whitespace!


katiestahl:
going to go a different route and point to readthedocs instead

Issue #331, opened 2016-06-22
hgvs doesn't support uncertain offsets


reece:
Originally reported by: **Reece Hart (Bitbucket: [reece](http://bitbucket.org/reece), GitHub: [reece](http://github.com/reece))**

----------------------------------------

The parser should accept uncertain offsets 

```
#!python

>>> hp.parse_hgvs_variant('NM_024675.3:c.-200-?_*297+?del')
Traceback (most recent call last):
  File "<ipython-input-2-8a4387bb8595>", line 1, in <module>
    hp.parse_hgvs_variant('NM_024675.3:c.-200-?_*297+?del')
  File "/home/reece/projects/biocommons/hgvs/hgvs/parser.py", line 108, in rule_fxn
    reason=exc.formatReason()))
HGVSParseError: NM_024675.3:c.-200-?_*297+?del: char 19: expected a digit

>>> hp.parse_c_pos("56+?")
Traceback (most recent call last):
  File "<ipython-input-6-b546f1972089>", line 1, in <module>
    hp.parse_c_pos("56+?")
  File "/home/reece/projects/biocommons/hgvs/hgvs/parser.py", line 108, in rule_fxn
    reason=exc.formatReason()))
HGVSParseError: 56+?: char 4: expected EOF

>>> hp.parse_c_pos("?+2")
BaseOffsetPosition(base=None, offset=2, datum=1, uncertain=False)

```

The parser and models currently don't support uncertain offsets. They do support uncertain base positions.

----------------------------------------
- Bitbucket: https://bitbucket.org/biocommons/hgvs/issue/331

Issue #225, opened 2015-04-22
support uncertain ranges


reece:
Originally reported by: **Reece Hart (Bitbucket: [reece](http://bitbucket.org/reece), GitHub: [reece](http://github.com/reece))**

----------------------------------------

Indefinite positions such as g.123_(450_456)del are not currently supported

----------------------------------------
- Bitbucket: https://bitbucket.org/biocommons/hgvs/issue/225

reece:
Related: #331

davmlaw:
Here's some example ranges from ClinVar

| ClinVar Variation ID | SV type | g.HGVS | c.HGVS |
| -------- | -------- | -------- | -------- |
| [11692](https://www.ncbi.nlm.nih.gov/clinvar/variation/11692/) | DEL | ```NC_000023.11:g.(133661675_133661730)_(133661850_133661926)del``` | ```NM_004484.3(GPC3):c.(1293_1293)-76_(1413_1413)del``` |
| [31562](https://www.ncbi.nlm.nih.gov/clinvar/variation/31562/) | DEL | ```NC_000007.14:g.(16361566_16366648)_(16369693_16391969)del``` | ```NM_001101426.3(CRPPA):c.(535_684)+6399_(535_684)+14526del``` |
| [425669](https://www.ncbi.nlm.nih.gov/clinvar/variation/425669/) | DEL | ```NC_000002.12:g.(?_202376935)_(202377551_202464808)del``` | ```NM_001204.6(BMPR2):c.(?_-540)_(76+1_77-1)del``` |
| [425698](https://www.ncbi.nlm.nih.gov/clinvar/variation/425698/) | DEL | ```NC_000002.12:g.(202377551_202464808)_(202559947_?)del``` | ```NM_001204.6(BMPR2):c.(76+1_77-1)_(*1_?)del``` |
| [220591](https://www.ncbi.nlm.nih.gov/clinvar/variation/220591/) | DEL | ```NC_000017.11:g.(?_58709859)_(58734342_?)del``` | ```NM_058216.2(RAD51C):c.706-?_*120del``` | 
| [251062](https://www.ncbi.nlm.nih.gov/clinvar/variation/251062/) | DUP | ```NC_000019.9:g.(11211022_11213339)_(11217364_11218067)dup```   | ```NM_000527.5:c.(190+1_191-1)_(817+1_818-1)dup``` |
[...]

PR #699, opened 2023-09-20
Uncertain ranges

theferrit32:


theferrit32:
Haven't added for `c.` yet. The `hgvs.pymeta` changes are likely similar but will need to add tests.

korikuzma:
@biocommons/hgvs-maintainers Would we be able to get feedback on this PR?

andreasprlic:
It would be great to get this PR in. Can we just make sure the "uncertain" field on BaseOffsetPosition or SimplePosition gets set correctly (and tested in the unit test)?

Thanks!

tests/test_hgvs_sequencevariant.py
```diff
@@ -97,6 +97,30 @@ def test_format(self):
...
+        hp = hgvs.parser.Parser()
+
+        vs = "NC_000005.9:g.(90136803_90144453)_(90159675_90261231)dup"
+        v = hp.parse(vs)
+        self.assertEqual(vs, str(v))
+        self.assertEqual(v.posedit.pos.start.start.base, 90136803)
+        self.assertEqual(v.posedit.pos.start.end.base, 90144453)
+        self.assertEqual(v.posedit.pos.end.start.base, 90159675)
+        self.assertEqual(v.posedit.pos.end.end.base, 90261231)
+        self.assertEqual(type(v.posedit.edit).__name__, "Dup")
```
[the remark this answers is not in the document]
andreasprlic: Merged #713 in to unblock CI and added the tests I thought were missing

andreasprlic (approved):
LGTM

andreasprlic:
Playing around with this branch some more I realized that g_to_c projection was not yet working. As such I had a go at enabling this. Since the behavior of this projection is somewhat undefined, the current approach picks the inner (=confident) interval of the uncertain range and uses that for the .c projection. The resulting hgvs_c string is then "confident" and not "uncertain" any longer.

andreasprlic:
Can we work on this one on Monday during our dev session?

andreasprlic:
As discussed on Monday, here a screenshot of `NC_000009.11:g.108337304_108337428del`. If we would represent this as a closely related, but imprecise event `NC_000009.11:g.(?_108337304)_(108337428_?)del` , how would we like the hgvs_c to be?

Would we want g_to_c to resul in `NM_001079802.1:c.-10_105+10del` for both the precise as well as imprecise event? If we were to map this to `NM_001079802.1:c.(?_-10)_(105+10_?)del`, do we want this to be comparable with the precise form in some form?

Screenshot 2024-02-14 at 4 13 40 PM

reece:
I believe that the c. uncertainty should match the g. uncertainty. That is, my preference is

`NC_000009.11:g.(?_108337304)_(108337428_?)del` ⇔ `NM_001079802.1:c.(?_-10)_(105+10_?)del`

In principle, I don't see any problem with a partially uncertain event, although I don't see how it would apply in this case.

reece (changes requested):
An `x_interval` in this area of code is a definite (`def_`) interval or uncertain interval.
An uncertain interval is merely a definite interval in parentheses.
So, `def_g_uncertain_interval` doesn't make sense to me. What am I missing?

Also, if accepted, I think we'd want this interval form to apply to the other kinds of intervals.

andreasprlic:
Added imprecise hgvs_c strings, but it is not 100% there yet, since 

`NC_000009.11:g.(?_108337304)_(108337428_?)del` ⇔ `NM_001079802.1:c.(-10)_(105+10)del` at the moment. We want this to be `NM_001079802.1:c.(?_-10)_(105+10_?)del`. The problem is that the -10 comes out as a BaseOffsetPosition currently, when we prob would want it an Interval where the start has no "base" set. That prob. requires some more modifications on `g_to_n` in the alignmentmapper.

(see otherwise working cases in test_uncertain_projection_confidence)

reece:
Thanks for working this issue, Andreas.

I converted this PR to a draft to make it clear that it was a WIP.

larrybabb:
@andreasprlic thank you for working on this. This is a fairly important issue for vrs-Python and the various datasets we are trying to transform from hgvs to vrs. Do you have any sense of when this work will be completed or if it could be handed off to another developer efficiently? No pressure but any information will assist our planning.

andreasprlic:
@larrybabb Thanks for calling this out. So far we have been treating this one as a "nice to have", not a "must have". 

At this point I believe g_to_c for imprecise events is working. Right now I am looking into the other direction, c_to_g, which will require some "grammar" modifications in how to parse the hgvs-strings. Perhaps it is best if I push what I got so far up, then somebody else who has worked on the parser previously could try to get imprecise hgvs_c parsing to work? @theferrit32 perhaps? :-)

andreasprlic:
I  pushed my current version of the code.

Note: 
- unit tests for imprecise g_to_c are passing. 
- unit test for partially precise g_to_c is passing too.
- not passing yet: imprecise c_to_g.

andreasprlic:
@larrybabb for your use case, do you need the support in the direction `g_to_c`, or do you also need `c_to_g`?  I do wonder if the vast majority of use cases are addressed by this PR as it is, and the missing parsing of imprecise c_to_g would be just useful for rare edge cases (and could be done via a future PR).

larrybabb:
@andreasprlic at this point in time I do not need either the `g_to_c` or the `c_to_g` support. But if I look ahead a bit I think I might like to have the `g_to_c` support down the road.  I do not foresee needing the `c_to_g` on these ambiguous endpoints.

src/hgvs/_data/hgvs.pymeta
```diff
@@ -122,7 +122,7 @@ pro_ident  = '='                             -> hgvs.edit.AARefAlt(ref='',alt=''
 
 # potentially indefinite/uncertain intervals
 c_interval = def_c_interval | '(' def_c_interval:iv ')' -> iv._set_uncertain()
-g_interval = def_g_interval | '(' def_g_interval:iv ')' -> iv._set_uncertain()
+g_interval = uncertain_g_interval:iv | ('(' def_g_interval:iv ')' -> iv._set_uncertain()) | def_g_interval
```
reece, line 125: I'm confused by this. The second alternative parse (with parentheses) is for an uncertain g_interval.

reece (changes requested):
@andreasprlic @theferrit32 @larrybabb: I figured out my discomfort with the changes in around hgvs.pymeta:125. At the very least, I'd like to discuss this further because I think the current change breaks important parity across variant types.

We currently have coordinates like:
- `5` (`pos` in the grammar)
- `5_5` (`def_x_interval`)
- `(5_5)` (`def_x_interval` with Interval.uncertain = True)

And now we're adding `(5_5)_(5_5)` for genomic sequences only. I think there are two problems with this. 1) We should have parallel constructs for other variant types, 2) `uncertain_g_interval` is a confusing name (to me) because it can be confused with the `(5_5)` form above.

It seems to me that we should be able to refactor this into atomic positions (as is) and a single interval with starts and ends that can be either positions or, now also intervals.

I haven't thought this through fully, but I think this could be implemented by creating a `g_pos_or_interval` rule, then using that in line 132 in lieu of the g_pos:start and :end.

Thanks for your patience. I'm not trying to drive you crazy, but the parity of these rules is part of what makes the hgvs package useful in structuring HGVS variant descriptions, so I'm reluctant to toss this parity across rules without a compelling reason.

reece:
Per collective coding  meeting discussion on 2024-09-09, here's the path forward:

- Rework grammar changes to preserve analogous grammar across sequence types for def_g_interval, etc.
- Update g-to-c projections to support transformations like those shown in https://github.com/biocommons/hgvs/issues/225#issuecomment-1661318086 
- Ensure tests for each of these

andreasprlic:
I updated this PR with the following fixes:
- small parser improvement with uncertain_g_interval| def_g_interval
- incorporated unit tests for examples provided in #225 
- The test also demonstrates how the current behavior for imprecise ranges works. During g_to_c we lose the outer-confidence interval, and only maintain the inner confidence interval.

andreasprlic:
Note:  some of the provided c. examples in #225 are most likely not valid hgvs: e.g. I think this is not valid: `NM_000304.3:c.-34-?_*1140dup1657`. In the current form of this PR, after g_to_c this becomes: `NM_000304.3:c.(?_-34)_(*1140_?)dup'.

reece (commented):
It seems to me that this code still isn't quite ready, or we need to be clearer about the end goal.

I'm mostly over the minor issues I had with grammar changes. (But see comments.)

The bigger issue is that I think we haven't actually solved #225 (or #331). Should we narrow the goal for this PR? @andreasprlic made relevant comments in the discussion thread as well.

reece:
> Note: some of the provided c. examples in #225 are most likely not valid hgvs: e.g. I think this is not valid: `NM_000304.3:c.-34-?_*1140dup1657`. In the current form of this PR, after g_to_c this becomes: `NM_000304.3:c.(?_-34)_(*1140_?)dup`.

Clinvar genomic  NC_000017.10:g.(?_15133094)_(15164078_?)dup
hgvs:            NM_000304.3                 :c. (?_-34)_(*1140_?)  dup
ClinVar has: NM_000304.3(PMP22)  :c. -34-? _ *1140        dup1657

>>> v = parse("NM_000304.3(PMP22):c.-34-?_*1140dup1657")
HGVSParseError: NM_000304.3(PMP22):c.-34-?_*1140dup1657: char 25: expected a digit

-34-? is not be parsed by hgvs currently because `offset = snum | (-> 0)` (hgvs.pymeta:202)

from tests in #225, clinvar [251062](https://www.ncbi.nlm.nih.gov/clinvar/variation/251062/)
clinvar g                "NC_000019.9:g.(11211022_11213339)_(11217364_11218067)dup",
clinvar c                "NM_000527.5:c.(190+1_191-1)_(817+1_818-1)dup",
hgvs g                "NC_000019.9:g.(?_11213339)_(11217364_?)dup",
hgvs c                "NM_000527.5:c.(?_191-1)_(817+1_?)dup",

On current #699 branch:
>>> var_g1 = parse("NC_000019.9:g.(11211022_11213339)_(11217364_11218067)dup")
>>> var_g1
SequenceVariant(ac=NC_000019.9, type=g, posedit=(11211022_11213339)_(11217364_11218067)dup, gene=None)
>>> var_c1 = am38.g_to_c(var_g1, "NM_000527.5")
>>> var_c1
SequenceVariant(ac=NM_000527.5, type=c, posedit=(?_191-1)_(817+1_?)dup, gene=None)
>>> var_g2 = am38.c_to_g(var_c1)
Traceback (most recent call last):
  Cell In[16], line 1
    var_g2 = am38.c_to_g(var_c1)
  File ~/projects/biocommons/hgvs/src/hgvs/assemblymapper.py:131 in c_to_g
    return self._maybe_normalize(var_out)
  File ~/projects/biocommons/hgvs/src/hgvs/assemblymapper.py:265 in _maybe_normalize
    return self._norm.normalize(var)
  File ~/projects/biocommons/hgvs/src/hgvs/normalizer.py:119 in normalize
    if var.posedit.pos.start.base < 0 or not is_valid_pos(var.ac, var.posedit.pos.end.base):
AttributeError: 'Interval' object has no attribute 'base'


Action items:
- Retain precision of inner and outer boundaries in g_to_c and c_to_g; see example above
- Add support for ? in offset; consider putting in separate issue
- 

                          (190,None)
                        (190+1,None)
         (190+1       _        191-1)
(190+1 _ 191-1) _ (817+1 _ 818-1)

reece:
@andreasprlic @theferrit32 I did a bunch of surgery on project config diffs between main and this branch. They're now in sync. Tests pass, including on github (i.e., the cache is intact)

Please destroy your venv and rebuild with `make devready` to test.

davmlaw:
It would be good if you could summarise the current state of things and what needs to be done.

From my quick skim - I'm worried about the silent conversion between types eg g_to_n. The HGVS spec doesn't list how conversions should be done, so we are on our own.

In  ```g_to_n``` - comment is "in case of uncertain ranges, we fall back to the inner (more confident) interval"

This is a valid choice, but I argue it's not the only one.

I would argue:

* We can't convert things silently - the only possible default is to return "p.?" or "c.?"
* We should offer a SequenceVariant convenience method to return a new SequenceVariant object representing uncertain inner interval, so clients can call g_to_n with that explicitly
* We could offer an option in mapper to do this (eg ```uncertain_conversion='inner'``` but better name than I can think of now) which would do your current behavior

src/hgvs/alignmentmapper.py
```diff
@@ -312,6 +850,200 @@ def g_interval_is_inbounds(self, ival):
...
+            else:
+                grs_orig_end = MAX_CHR_SIZE + 1
+
+            # Sort the four values from smallest to largest
+            all_positions = [gre_orig_start, gre_orig_end, grs_orig_start, grs_orig_end]
+            all_positions.sort()
+
+            # Convert sentinel values back to None
+            all_positions = [
+                None if x == -1 or x >= MAX_CHR_SIZE else x for x in all_positions
```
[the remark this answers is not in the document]
andreasprlic: If that does not work use 9999999...

src/hgvs/alignmentmapper.py
```diff
@@ -150,72 +162,493 @@ def __str__(self):
...
+        offsets[0], offsets[1] = -orig_offsets[1], -orig_offsets[0]
+        offsets[2], offsets[3] = -orig_offsets[3], -orig_offsets[2]
+
+        # Swap cigar operations
+        cigars[0], cigars[1] = orig_cigars[1], orig_cigars[0]
+        cigars[2], cigars[3] = orig_cigars[3], orig_cigars[2]
+
+        return positions, offsets, cigars
+
+    def _transform_two_positions_for_reverse_strand(self, pos1, pos2, offset1, offset2):
```
andreasprlic, line 249: take this out, no longer needed it seems.

src/hgvs/location.py
```diff
@@ -61,19 +65,38 @@ def validate(self):
...
+
+        if lhs.base is None and rhs.base is None:
+            raise HGVSUnsupportedOperationError(
+                "Cannot compare two positions without bases"
+            )
+
+        # imprecise positions can be on both sides of an interval
+        # This is weird, but because an unknown breakpoint can be expressed on both sides
+        # with a ? character we need to support that both options are true
+        if lhs.base is None or rhs.base is None:
```
andreasprlic, line 97: test: if left is None -> True, if right is None ->False

Issue #712, opened 2023-12-11
make develop not working


andreasprlic:
**Describe the bug**

Trying to update to the develop setup for the current main branch is broken.

**To Reproduce**
```shell
python3 -m venv
source venv/bin/activate
make develop
```
results in an error message complaining about the dev dependencies that are defined in setup.cfg

**Expected behavior**
It should just work

**Additional context**
Will have a fix shortly

andreasprlic:
fixed by #713

PR #713, opened 2023-12-11
fix(#712): This fixes the broken make develop install

andreasprlic:
This PR fixes the broken make develop, by making sure that the optional dev dependencies are indicated as optional in pyproject.toml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ keywords = [
     "hgvs",
     "variation"
     ]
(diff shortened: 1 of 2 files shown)


andreasprlic:
... as we can see from the broken CI, apparently `make test` is also broken ATM

reece (approved):
Thanks for picking this up. The enforcement of this change has created lots of headaches in many packages that I use. :-(

PR #716, opened 2024-01-04
Add variant region condition for dups ending in UTR

### LINKED ISSUE #715: c_to_p for dups that start in transcript but end in UTR
Any duplications with an end position at or past the stop codon should be classified as 3'UTR regardless of start position. Currently mapping `NM_153223.3:c.2959_*1dup` yields `NP_694955.2:p.(Met1?)`. We believe that should map to `NP_694955.2:p.?` because all other variants in the UTR map to `p.?`. 

```
In [1]: var_c = parse('NM_153223.3:c.2959_*1dup')

In [2]: var_p = c_to_p(var_c)

In [3]: var_p
Out[3]: SequenceVariant(ac=NP_694955.2, type=p, posedit=Met1?, gene=None)

In [4]: str(var_p)
Out[4]: 'NP_694955.2:p.Met1?'
```

We expect this to result in `NP_694955.2:p.?` instead.

b0d0nne11:
Fixes #715.

Classify `ins` or `dup` variants as 3'UTR if the end position is in the UTR region regardless of the start position.


b0d0nne11:
Including `ins` here as per my comment on the related issue.

reece (commented):
Please see comment in #715

b0d0nne11:
Rebased on main

andreasprlic (approved):
Thanks, these changes look good to me.

PR #719, opened 2024-01-23
Fix ins or dups where splice region is preserved

### LINKED ISSUE #714: c_to_p at intron/exon boundary where splice region is preserved
We have a number of ins or dup variants at the intron/exon or exon/intron boundary that return no protein change that we believe should be treated as coding because the splice site & region remain completely intact. This is related to issue #655. There we handled variants that occur at boundary but here we're concerned with variants that span the boundary. We chose to treat these as separate issues because we can't rely only on the offset positions for these variants so our approach needs to be different.

For example:
```
In [1]: var_c = parse('NM_004119.2:c.1837+21_1837+22insCGAGAGAATATGAATATGATCTCAAATGGGAGTTTCCAAGAGAAAATTTAGAGTTTGGTAAGAATGGAATGTGCCAAA')
In [2]: c_to_p(var_c)
Out[2]: SequenceVariant(ac=NP_004110.2, type=p, posedit=None, gene=None)
```
We expect this to map to `NP_004110.2:p.(Lys614_Val615insAsnGlyMetCysGlnThrArgGluTyrGluTyrAspLeuLysTrpGluPheProArgGluAsnLeuGluPheGlyLys)`

To fix, when calculating c_to_p if you end up with no protein change try shifting the var_g (negative_normalizer if strand==1, normalizer if strand==-1) and then sending the shifted mutation through the g_to_c and c_to_p to see if it gets a protein change. If it gets a protein change, report this change as hgvs_p.

For example:
```
In [3]: var_c = parse('NM_004119.2:c.1837+21_1837+22insCGAGAGAATATGAATATGATCTCAAATGGGAGTTTCCAAGAGAAAATTTAGAGTTTGGTAAGAATGGAATGTGCCAAA')
In [4]: import hgvs
In [5]: normalizer = hgvs.normalizer.Normalizer(hdp, shuffle_direction=3)
In [6]: var_g = normalizer.normalize(c_to_g(var_c))
In [7]: var_c = g_to_c(var_g, var_c.ac)
In [8]: c_to_p(var_c)
Out[8]: SequenceVariant(ac=NP_004110.2, type=p, posedit=(Lys614_Val615insAsnGlyMetCysGlnThrArgGluTyrGluTyrAspLeuLysTrpGluPheProArgGluAsnLeuGluPheGlyLys), gene=None)
```

b0d0nne11:
Fixes #714.

Fixes ins or dup variants spanning the intron/exon or exon/intron boundary where the splice site & region remain completely intact.


b0d0nne11:
We found some examples of duplications where the original logic here didn't shift the variant far enough to to get the expected result. It turned out that for these variants it's not possible to write the shifted version as a duplication since in a duplication the alt will always follow the ref. I've added some logic to rewrite these shifted variants as insertions before attempting to map them back to `var_p`s and added tests to include these cases.

gostachowiak:
High-level explanation of this pull request: 
- HGVS nomenclature has the 3' shifting rule.  so all cdots and pdots are shifted to the right
- However, 3' shifting is arbitrary and a necessary evil for nomenclature purposes.  But biology doesn't care about the 3' shifting rule
- Consider an example:  positive strand gene, first 8 bases of the intron is duplicated
- cdot would be +1_+8dup
- currently no pdot would be calculated because both positions have an offset
- but now consider the biology-- after the duplication, there are 2 splice sites on the left side of the intron.  Which is more likely to be used for splicing? 
- We obviously can't know for sure, but it seems to me that the most logical assumption is that it will use the "inner" splice site for splicing, leaving the extra inserted material within the coding sequence, resulting in a frameshift pdot.  Why is this the best assumption? 
-- When that splice site is used, the entire intronic sequence is totally intact
-- Seems "safer" to calculate a pdot for this to rescue the variant-- otherwise downstream applications will most likely be filtering out this variant because it's an insertion after the 8th position in the intron
- To handle this type of situation in the most general way possible, this is the approach: 
-- calculate pdot the normal way.  
-- if empty, shift the cdot in the REVERSE direction, and calculate the pdot again.  if you get a result, use it
-- basically, if you can get a pdot with either forward or reverse shifting, that means the entire intron is intact and we should bring in the pdot
- note that the 3' shifting rule is still respected both for the cdot and pdot nomenclature
-- reverse shifting is only used as a tool when calculating pdot from cdot, which does not violate HGVS nomenclature
-

gostachowiak:
I also wanted to mention that we discovered this because some fraction of FLT3 ITDs currently get missed when using the hgvs package (including one added to the unit tests).  So it is a high impact issue.

b0d0nne11:
Commenting to remove the stale label...

b0d0nne11:
@reece or @ahwagner can you remove the stale label here? We would still like to get this merged is possible. Thanks!

b0d0nne11:
We found a case where trying to map the shifted variant causes an `HGVSInvalidVariantError`. I've added logic to handle this and a test case. The variant is `NM_182758.2:c.2953-31_2953-26dup`. As part of the shifting procedure, mapping this to the `g` type yielded an unexpected transformation to `NC_000015.9:g.53815545_53815550delinsC` that caused problems with later steps. I'm simply handling the error here since we don't want to consider invalid variants.

b0d0nne11:
Also rebased on main

b0d0nne11:
Made a few non-functional edits:
- Rebased on main
- Moved tests to an isolated test file
- Squashed some commits

ahwagner:
@b0d0nne11 is this PR ready for review?

b0d0nne11:
Hi @ahwagner, I've been focusing on https://github.com/biocommons/hgvs/pull/748 recently. I think it would be better to get that merged first and then circle back around to this. We definitely still want to get this merged but I think some of the decisions we make there might affect this PR also.

korikuzma (commented):
@b0d0nne11 said they're waiting until #748 is complete. Re-request review once ready.

b0d0nne11:
@korikuzma @ahwagner I think this is ready for a look now. I made a couple of changes beyond just refactoring. I added a configuration parameter to control whether to shift these variants into the exon or the intron. I also added flags to `posedit` to show that the variant is either on the boundary or was shifted. This was requested by one of our clients since it's difficult to tell which variants where shifted after the fact. I wasn't sure that `posedit` was the correct place to put those. I think a case could be made to put them on the `var` or `pos` object instead. I would be happy to move them if you prefer.

b0d0nne11:
Hey @reece, thanks! I rebased on master and took care of the merge conflicts. I'll also be out next week so no rush. Whenever you have time would be great.

src/hgvs/_data/defaults.ini
```diff
@@ -7,6 +7,8 @@ normalize = True
 prevalidation_level = EXTRINSIC
 replace_reference = True
 ins_at_boundary_is_intronic = True
+shift_over_boundary = False
```
reece, line 10: @b0d0nne11: Would you please add a short comment above each of these describing what they do? Also note that False + DEFAULT correspond to hgvs 1.x code behavior and the team's interpretation of HGVS Nomenclature.

src/hgvs/utils/altseqbuilder.py
```diff
@@ -176,7 +176,7 @@ def build_altseq(self):
 
         return alt_data
 
-    def _get_variant_region(self):
+    def get_variant_region(self):
```
reece, line 179: The complexity of these cases warrants explicit tests. Would you please implement those? I *think* that it might be enough to roll these cases into the `cases` block of `test_714.py` with an explicit call to `get_variant_region()`.

tests/data/gcp/real.tsv
```diff
@@ -50,14 +50,12 @@ ID00048	NC_000010.10:g.89692922T>C	NM_000314.4:c.406T>C	NP_000305.3:p.(Cys136Arg
 ID00049	NC_000010.10:g.89692921dupA	NM_000314.4:c.405dupA	NP_000305.3:p.(Cys136Metfs*44)
 ID00050	NC_000010.10:g.89692923_89692939delGTGCATATTTATTACAT	NM_000314.4:c.407_423delGTGCATATTTATTACAT	NP_000305.3:p.(Cys136Serfs*38)
 ID00051	NC_000010.10:g.89712015C>A	NM_000314.4:c.633C>A	NP_000305.3:p.(Cys211*)
-ID00052	NC_000010.10:g.89685314dupT	NM_000314.4:c.209dupT	NP_000305.3:p.?
```
reece, line 53: Please comment why these tests were removed. It scares me to remove tests because it implies that the tests were wrong and therefore that the code was passing something that it should have. These particular tests were contributed by some extremely experienced clinical geneticists at Invitae (in ~2015). 

To be clear, we should remove them if they're wrong. I just want to know that rationale explicitly. And, while I can see the two deleted tests might be questionable, the rationale that I would use would implicate most or all of these p.? tests, not just these two.

src/hgvs/variantmapper.py
```diff
@@ -700,6 +730,61 @@ def _update_gene_symbol(self, var, symbol):
         var.gene = symbol
         return var
 
+    def _var_c_shifts(self, var_c, alt_ac, alt_aln_method):
```
reece, line 733: Please add tests for the _var_c_shifts and _var_g_shifts methods.

reece (changes requested):
@b0d0nne11- This is a great contribution! Thank you! I especially like the shift functions to explore alternative representations. And I appreciate your care in enabling users to opt into new behavior without impinging on existing functionality.

We're working toward a 2.0 release in December. I'd love to include this feature.

src/hgvs/variantmapper.py
```diff
@@ -700,6 +730,61 @@ def _update_gene_symbol(self, var, symbol):
...
+                yield shifted_var_g
+            except (HGVSInvalidVariantError, HGVSInvalidIntervalError, HGVSUnsupportedOperationError):
+                pass
+
+    def _var_g_shift_with_rewrite(self, var_g, shuffle_direction, strand, alt_aln_method):
+        """Attempt to shift a variant all the way left or right. Rewrite
+        duplications as insertions so that the variant is shifted farther
+        than would normally be possible using the HGVS notation."""
+        var_g = copy.deepcopy(var_g)
+        normalizer = hgvs.normalizer.Normalizer(
```
andreasprlic, line 764: rather than creating this here on every call of `_var_g_shift_with_rewrite`, should we move this (perhaps even as an optional argument) into the `VariantMapper.__init__()`?

src/hgvs/variantmapper.py
```diff
@@ -700,6 +730,61 @@ def _update_gene_symbol(self, var, symbol):
...
+            self.hdp, alt_aln_method=alt_aln_method, validate=False, shuffle_direction=shuffle_direction
+        )
+        var_g = normalizer.normalize(var_g)
```
andreasprlic, line 767: I did not test this, but I suspect that calling normalize() is slow. What would happen if we would not call the normalizer here?

andreasprlic:
Overall I am very happy with this PR, and I would like to approve it, once Reece's comments have been addressed (mostly about testing). This is an improvement that shows great attention to detail and that solves an important edge case in a way that can be configured (and the default is compliant with HGVS nomenclature). Thanks for putting in all this effort! This is greatly appreciated.

andreasprlic:
I am interested in getting this PR fixed and merged. As such I took a stab at it here:
#821

reece:
Closing as superseded by #821, which arrived at the same fix independently while resolving conflicts against `main`.

Evidence:
- `tests/issues/test_714.py` (the test encoding this fix's expected behavior) is byte-identical between the two PR branches.
- `tests/support/mock_input_source.py` is also byte-identical.
- `src/hgvs/_data/defaults.ini` and `src/hgvs/enums.py` (the `ShiftOverBoundaryPreference` enum and its config default) are identical between both branches.
- #821's own history contains an explicit merge commit stating it merged `ShiftOverBoundaryPreference` and `HGVSInvalidIntervalError` imports "from PR #719."
- The one apparent gap — this branch's `AssemblyMapper._var_c_shifts()` override and its constructor params — isn't missing functionality in #821, just a different architecture: #821 resolves `alt_ac` from `tx_ac` once at the top-level `AssemblyMapper.c_to_g`/`t_to_g`/etc. methods and threads it down to `VariantMapper._var_c_shifts` directly, rather than re-resolving it inside a per-shift override.

No changes from this branch are missing in #821.

Issue #704, opened 2023-10-07
Normalizer constructor doesn't pass the `validate` param to the VariantMapper constructor


mihaitodor:
When initialising an `AssemblyMapper`, a new `Normalizer` is initialised [here](https://github.com/biocommons/hgvs/blob/c69e03fa8b84a00c360e72f05de0e83177598cac/src/hgvs/assemblymapper.py#L91-L93). However, the `Normaliser` initialisation instantiates a new `VariantMapper` [here](https://github.com/biocommons/hgvs/blob/c69e03fa8b84a00c360e72f05de0e83177598cac/src/hgvs/normalizer.py#L58), but it doesn't set the `prevalidation_level` field for this `VariantMapper`, which means that it will fall back to reading the global config [here](https://github.com/biocommons/hgvs/blob/c69e03fa8b84a00c360e72f05de0e83177598cac/src/hgvs/variantmapper.py#L74).

I'm looking to avoid making calls to seqfetcher for transcripts and I bumped into this issue.

davmlaw:
Fixed w/Pull request

PR #725, opened 2024-02-07
fix: update gene symbol when it is `None` and `AssemblyMapper` normalizer

markgene:
Two fixes:

1. The `_update_gene_symbol` of `VariantMapper` class did not update symbol when it is `None`. 

2. The normalizer of `AssemblyMapper` uses a `VariantMapper` configured inconsistently from `AssemblyMapper` itself. The `AssemblyMapper._norm` is a `Normalizer` object. `Normalizer` has a field `vm` of `VariantMapper` class. In the current code, when initializing an `AssemblyMapper` object with the argument of a different value from `global_config` (e.g. `add_gene_symbol=True`, the `vm` will be created using `global_config`, rather than the parameters passed to `AssemblyMapper`. This is confusing, because the mappers are configured differently and can bring unexpected behaviors. 

The fix passes a `VariantMapper` object initialized using the same arguments as `AssemblyMapper` object.

### DIFF
--- a/src/hgvs/assemblymapper.py
+++ b/src/hgvs/assemblymapper.py
@@ -88,8 +88,11 @@ class AssemblyMapper(VariantMapper):
         self.in_par_assume = in_par_assume
         self._norm = None
         if self.normalize:
+            vm = VariantMapper(hdp=hdp, replace_reference=replace_reference, 
+                               prevalidation_level=prevalidation_level, 
+                               add_gene_symbol=add_gene_symbol)
             self._norm = hgvs.normalizer.Normalizer(
-                hdp, alt_aln_method=alt_aln_method, validate=False
+                hdp, alt_aln_method=alt_aln_method, validate=False, variantmapper=vm
             )
         self._assembly_map = {
             k: v for k, v in hdp.get_assembly_map(self.assembly_name).items() if k.startswith("NC_")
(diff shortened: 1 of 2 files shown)


davmlaw:
1. **_update_gene_symbol**

```
if not symbol:
```

Already handles symbol being None

I verified that the existing code handled None:

```
In [16]: var_c.gene = None

In [17]: vm._update_gene_symbol(var_c, "test_gene_symbol")  # existing code
Out[17]: SequenceVariant(ac=NM_002386.3, type=c, posedit=-5_1del, gene=test_gene_symbol)

In [18]: var_c.gene
Out[18]: 'test_gene_symbol'
```

Next time - it's good to include a test that fails on the old code, and then passes on the new code (eg an automated test doing what I did above). I think if you did this, you would have discovered that the code change was unnecessary

2. Given an AssemblyMapper is a VariantMapper, could you pass ```variantmapper=self```? That would future proof the code so that we don't need to update any new params we add

Also - is this a fix for issue #704 - if so, could you please link issue in pull request? Thanks

korikuzma:
@markgene I agree with @davmlaw 's comments.

korikuzma (commented):
@markgene I agree with @davmlaw 's comments.

davmlaw:
@markgene - thanks for your contribution, I have taken your change, removed the gene parts, and added a unit test - submitted as a new pull request: https://github.com/biocommons/hgvs/pull/746

Issue #178, opened 2014-06-11
use abc for Edit class to ensure method definition


reece:
Originally reported by: **Reece Hart (Bitbucket: [reece](http://bitbucket.org/reece), GitHub: [reece](http://github.com/reece))**

----------------------------------------

Related to: #178, #301, #347

Issue #175, in which the type method was implemented only for some subclasses of the Edit class, suggests that we use the abc package and better subclassing to improve consistency.

It's unclear whether/how this interacts with recordtype used by current classes.

----------------------------------------
- Bitbucket: https://bitbucket.org/biocommons/hgvs/issue/178

davmlaw:
Fixed with pull request

PR #729, opened 2024-03-05
#178 - use abc for Edit class to ensure method definition

davmlaw:
Fix for issue #178 

While changing superclasses, I also removed Python2 objects, went with Python3 classes


korikuzma (commented):
This look good to me. It does appear that your branch is not up to date with the main branch anymore. Would you be able to update?

davmlaw:
@korikuzma - thanks for looking, have updated

davmlaw:
My other pull request was accepted... so updating from master again

Issue #730, opened 2024-03-08
For start loss AARefAlt.format always returns 3 letter amino acid even when single letter amino acid is configured 


jPleyte:
When working with a start loss variant and using the AssemblyMapper to obtain the p. genotype the return value is hard coded to return the tree letter amino acid value "Met1?" even when configured to return changes using single letter amino acids. 

**To Reproduce**
```py
am = hgvs.assemblymapper.AssemblyMapper(hdp, assembly_name=ASSEMBLY_VERSION, alt_aln_method='splign')

var_g = hgvs_parser.parse_hgvs_variant('NC_000016.9:g.89985662_89985667del')
var_c = am.g_to_c(var_g, str('NM_002386.3'))
var_p = am.c_to_p(var_c)
var_p_one_letter = var_p.format(conf={"p_3_letter": False})
var_p_three_letter = var_p.format(conf={"p_3_letter": True})

print(var_p_one_letter)    # NP_002377.4:p.Met1?
print(var_p_three_letter)  # NP_002377.4:p.Met1?
```

**Expected behavior**
When the SequenceVariant is configured to use single letter amino acids then the call to ``var_p.format`` should return a single letter amino acid representation of the change.

**Additional context**
The fix for this issue will be fairly simple.  In [edit.py](https://github.com/biocommons/hgvs/blob/main/src/hgvs/edit.py) we just need to check the configuration and return "p.Met?" or "p.M?".

PR #731, opened 2024-03-13
#730 - Fix single letter amino acid bug 

jPleyte:
In edit.py the ``AARefAlt.format`` function is hard-coded to return "Met1?" for start loss variants but it wasn't looking at the configuration map to determine if single letter format was requested. It was always returning the three letter format even when the single letter ("M1?") format was requested. 

* Made the fix to edit.py
* Created a unit test
* Updated cache 
* Formatted code


davmlaw:
Looks good to me (though I am not an official revewer), I confirmed the bug on old code, and that it fixes new code, and covered by test.

Also had a quick skim for other instances of bug and all other 3 letter AA are behind tests for p_3_letter

andreasprlic (approved):
Thanks, LGTM

PR #734, opened 2024-03-22
feat: update default data version config

### LINKED ISSUE #733: Update default config to uta v20210129b?
https://github.com/biocommons/hgvs/blob/67f610be7148bb47bac493abf40a0c0eb7a99deb/src/hgvs/_data/defaults.ini#L43

The default config is pointing to an outdated uta release. Would it be reasonable to bring this up to the most current available version?

jsstevenson:
close #733

Update default UTA version to `20210129b`

Did a search for other likely places where version values were hardcoded. Noticed that `defaults.ini` seems to be copy-pasted into the `config` section of the API reference, so replaced that with a [literalinclude](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) directive to reduce work in the future.

(I'm guessing I need to update some cassettes or something, so making this a draft for now)

### DIFF
--- a/tests/test_hgvs_dataproviders_uta.py
+++ b/tests/test_hgvs_dataproviders_uta.py
@@ -66,20 +66,20 @@ class UTA_Base:
 
     def test_get_tx_for_gene(self):
         tig = self.hdp.get_tx_for_gene("VHL")
-
+        
         expected_data = [
-            ("NM_000551.4",	"NC_000003.12"), # added in UTA at 2021-01-31 
-            ("NM_198156.2",	"NC_018914.2"),
-            ("ENST00000345392",	"NC_000003.11")
+            ("NM_001354723.1", "NC_000003.11"),
+            ("NM_198156.2", "NC_018914.2"),
+            ("ENST00000345392", "NC_000003.11"),
         ]
         found = 0
         for hgnc, cds_start_i, cds_end_i, tx_ac, alt_ac, alt_aln_method in tig:
             for t_tx_ac, t_alt_ac in expected_data:
                 if t_tx_ac == tx_ac and t_alt_ac == alt_ac:
                     found += 1
-        
+
         self.assertEqual(found, 3)
-        
+
     def test_get_tx_for_gene_invalid_gene(self):
         tig = self.hdp.get_tx_for_gene("GENE")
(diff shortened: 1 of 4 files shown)


reece:
@jsstevenson reported that tests failed with this change. Investigate because I think the tests should still work.

andreasprlic:
The reason why some of the unit tests fail is because we are jumping several UTA versions (from 2018 to late 2021) and new transcripts were loaded in the meanwhile. 

There is a [unit test for the get_tx_for_gene](https://github.com/biocommons/hgvs/blob/67f610be7148bb47bac493abf40a0c0eb7a99deb/src/hgvs/dataproviders/uta.py#L330) method for the VHL gene that expects a certain nr of transcripts in the response. The new database now contains more transcripts. I checked and these were loaded into UTA in 2020 and 2021. As such we should just fix that test. We should not test for an expected nr. of transcripts but rather test that a few specific tx_ac and alt_ac combinations are in the response from the query.

andreasprlic:
While I was at it, I fixed the problematic unit test...

tests/test_hgvs_dataproviders_uta.py
```diff
@@ -66,7 +66,19 @@ def test_get_tx_exons_invalid_alt_aln_method(self):
...
         tig = self.hdp.get_tx_for_gene("VHL")
-        self.assertEqual(16, len(tig))
+        
+        expected_data = [
+            ("NM_001354723.1", "NC_000003.11"),
+            ("NM_198156.2", "NC_018914.2"),
+            ("ENST00000345392", "NC_000003.11"),
+        ]
+        found = 0
+        for hgnc, cds_start_i, cds_end_i, tx_ac, alt_ac, alt_aln_method in tig:
```
korikuzma, line 76: hgnc, cds_start_i, cds_end_i, alt_aln_method aren’t used 
```suggestion
        for _, _, _, tx_ac, alt_ac, _ in tig:
```

korikuzma (approved):

jsstevenson:
@andreasprlic @korikuzma thanks for the assist! looks like we're good to merge now?

korikuzma (approved):

PR #737, opened 2024-05-21
chore: remove trailing whitespace

jsstevenson:
A lot of modern editors will auto-trim trailing whitespace upon save. Unfortunately, if there's a pesky space or two hanging around on some old file, then a new commit might create an unrelated edit on that line simply by making a functional change elsewhere in the file. This PR gets this out of the way for every `.py` file.


korikuzma (approved):
Not sure if you plan to do this change in other repos, but it might make sense to let the ruff format PR handle this

jsstevenson:
@korikuzma that's a good point -- fwiw, I checked every one of the major repos and only this + bioutils had any instances to fix

PR #740, opened 2024-06-05
style: format with Ruff

jsstevenson:


jsstevenson:
There are several open PRs right now -- I'd realllly prefer to get them merged before this, rather than forcing them to handle any conflicts created here.

Issue #741, opened 2024-06-10
Context view improvements


andreasprlic:
The hgvs library has a built in (text-based) visualization, which allows to build a view of the context of a variant with the alignment between the transcript and the reference genome. It can create representations similar to this:

```
                                              v                               NC_000010.10:g.64572045dupT
NC_000010.10 g 64572025 > ACTCAGGGAGTGATTTTTTTTCTCCATAATAAGGCAACCCA          > 64572065 NC_000010.10:g.64572045dupT
NC_000010.10 g 64572025 < TGAGTCCCTCACTAAAAAAAAGAGGTATTATTCCGTTGGGT          < 64572065 NC_000010.10:g.64572045dupT
                          |||||||||||||-|||||||||||||||||||||||||||          13=1D27=
NM_000399.3  n     2670 < TGAGTCCCTCACT-AAAAAAAGAGGTATTATTCCGTTGGGT          <     2709 NM_000399.3:n.2696dupA
NM_000399.3  c      902 <                                                    <      941 NM_000399.3:c.*928dupA
```

At the moment this visualization is flagged as "experimental". It also requires the uta_align package for re-aligning the sequences.

**Describe the solution you'd like**

It would be nice to expand on this and add a few more features:
- Show genome, transcript, protein sequences, plus something like a "ruler" to show the positions.
- Have a flexible windows size, that adjusts based on the size of a variant. 
- Exposes the data that is behind the view, so alternative renderings (perhaps SVG graphics?) could be built on top of this too
- Does not require the uta_align module, since we should have all the alignments already in UTA.
- nice to have: option to add some color to improve readability.
 
**Describe alternatives you've considered**

- I don't think there is anything quite like that yet for the hgvs library (besides what is already in context.py)

The question is mostly if we want to have better tooling around visualizing as part of the main hgvs module, or perhaps as a separate tool. Since we already have context.py, perhaps it fits into the main library.

jsstevenson:
Maybe this is a bit too out there, but for those publishing work to Jupyter notebooks (probably a lot of us), you can supply [special __repr__ methods](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display) that incorporate HTML/CSS (Pandas dataframes are probably the most popular example of this)

andreasprlic:
Actually that would be nice. Being able to show the context in a notebook would be a good feature to have.

reece:
@andreasprlic I think this issue is related to #742 , right?

PR #742, opened 2024-06-10
feat(#741): Pretty print

andreasprlic:
This (draft) PR is an attempt to add a context-visualization framework on top of the tooling that we already have in hgvs. Unless somebody has a better suggestion for a name, for now I called it "pretty_print".

Features:

1) It provides a [datacompiler](https://github.com/andreasprlic/hgvs/blob/pretty_print/src/hgvs/pretty/datacompiler.py) class that merges together all the data needed for the visualization. (no need for uta_align here). This could be used e.g. by a future effort to developer alternative visualisations such as SVG graphics, or hooked up on a web site.
2) For now there is only a [text-based rendering framework](https://github.com/andreasprlic/hgvs/tree/pretty_print/src/hgvs/pretty/renderer) that has been added here. It offers a set of elemental "renderer" objects that are responsible for providing the text for one line in the final text.
3) There are several unit tests that contain variants with various features. They are also the reason why this PR is still only in draft. The fixture that gets compiled for this unit test (tests/cache-py3.hdp) becomes really big and some suggestions for what to do about this would be helpful. Perhaps disable most of the tests? Or offer an ipython runbook with examples? Any feedback/suggestion is welcome here.
4) There's a configuration class where some parameters around the display can be modified
5) On a different thread we were discussing improvements around repeats. I wrote a basic repeat-detection script, and added the result of that as a FYI to the visualization here too, so we can see how well this works.

Examples:
 
 ```
In [1]: hgvs_g = "NC_000005.10:g.123346517_123346518insATTA"

In [2]: var_g = parse(hgvs_g)

In [3]: print(pretty.display(var_g))
hgvs_g    : NC_000005.10:g.123346517_123346518insATTA
hgvs_c    : NM_001166226.1:c.*1_*2insTAAT
hgvs_p    : NP_001159698.1:p.?
          :   123,346,500         123,346,520         123,346,540
chrom pos :   |    .    |    .    |    .    |    .    |    .
seq    -> : ATAAAGCTTTTCCAAATGTTATTAATTACTGGCATTGCTTTTTGCCAA
region    :                     |------|
tx seq <- : TATTTCGAAAAGGTTTACAATAATTAATGACCGTAACGAAAAACGGTT
tx pos    :  |    .    |    .   |   |    .    |    .    |
          :  *20       *10      *1  2880      2870      2860
aa seq <- :                      TerAsnSerAlaAsnSerLysAlaLeu
aa pos    :                         |||            ...
          :                         960
ref>alt   : ATTA[2]>ATTA[3]
```

Here also a screenshot how on the terminal the text can be color coded:

Screenshot 2024-06-30 at 22 53 19

### CODE REVIEW

**Comment by andreasprlic** [comment]
Note: this PR adds two new unit tests. They are pulling a lot of data from UTA and seqrepo, and as such they are slow. Should I tag them and exclude from the CI runs? Also, I did not want to upload an updated cache file for these tests, because the size of the test cache would have jumped from 1MB to ~600MB.

**Comment by davmlaw** [comment]
Wow, that's pretty neat and it's impressive you were able to do all of that without bringing in any dependencies (other than test framework)

From a very quick skim, it looks like you define your own data classes and do a lot of your own maths - I guess the worry would be if HGVS code is wrong, or yours is, or they're just different, then it would render differently than what's internally going on under the hood. Maybe there's no other way to do it, not sure - I'll try and block out some time next week to look over it

**Comment by andreasprlic** [comment]
Yes, the 'datacompiler' class is doing some calculations, but it is pretty much based around the internals of the hgvs library. The goal is to stitch together all the data around the context of a variant. The newly added "models" here are basic dataclasses to store the knowledge for one specific position in the final display.  The "renderer" classes are using the dataclasses to print the output.

One exception to this is the "repeats" class. This is derived from a conversation ( #113 ) about improving the support in the hgvs library for variants in repetitive regions. The current core of the library won't be easily able to be used for repeat detection in my opinion, but the fully-justified data calculations that are happening as part of the `datacompiler` make it relatively straightforward to identify basic repeats (that are not interrupted). I found it insightful for my own purposes to add this to the provided output in pretty_print and as such this ended up in here.

**Comment by reece** [comment]
From 8/26 maintainers' meeting: @andreasprlic: yes, please tag to exclude CI testing.

Also, we decided to keep the similar function https://github.com/biocommons/hgvs/blob/main/src/hgvs/utils/context.py for now. If someone cares, we can combine functionality later.

**Review by reece — CHANGES_REQUESTED** [review]
This is amazing. It's way better than src/hgvs/utils/context.py. Thank you.

I have a minor suggestion/request for discussion: Having hgvs.pretty_print and hgvs.pretty as siblings seems a bit confusing, and leads to (IMO) oddities like the renderer importing from a pretty_print.py.

Instead, I think a better structure might be to:
- put everything under pretty
- rename renderer to console
- put the color constants under console

I think that will create a better separate of the backend logic (immediately under pretty/) and the console renderer (under pretty/console/).

What say you?

**Comment by andreasprlic** [comment]
... with the most recent changes, I believe we are ready for another round of reviews.

File: src/hgvs/pretty/console/tx_pos.py
Comment (davmlaw) [inline src/hgvs/pretty/console/tx_pos.py:1]: RefSeq only... to handle Ensembl maybe you should look at something else?

Not sure if can test c_pos - or will have to look in transcript data to see if it has CDS start

Comment (andreasprlic) [reply src/hgvs/pretty/console/tx_pos.py:1]: fair point. That means this check should be in the data compiler and not here.

**Comment by andreasprlic** [comment]
In the latest revision, the default view is always to show transcripts 5'->3'. That means for reverse strand transcripts, we are flipping the orientation for the forward-strand. (behavior can be enabled/disabled with a config property)

Screenshot 2024-10-02 at 21 43 49

**Comment by ahwagner** [comment]
Removing stale for now and re-requesting review from prior reviewers. @reece and @davmlaw please take another look and confirm that your requested changes have been adequately addressed.

**Comment by davmlaw** [comment]
@andreasprlic  I've playing and I can see this being really useful!

A few nit-picks:

**Suggest changing PrettyPrint.__init__ to be initialised with AssemblyMapper**

Advantages are to support any future builds (eg T2T), or DataProviders that only support 1 assembly

At the moment, you instantiate 2 assembly mappers for hardcoded 37/38 and then only pick one, using ```default_assembly```. "default" is not really right - default is usually for a fall back, but this is the only choice ever used, so you only ever use 1 of the AssemblyMappers. If you passed in "GRCh37.p13" (or anything that doesn't match "GRCh37") it would give you the GRCh38 one

So - change would be to remove ```default_assembly``` and ```alt_aln_method``, replace with ```assembly_mapper: Optional[AssemblyMapper] = None```. Then if assembly_mapper is None, create one using 37/splign

This can then remove all the code that tests ```default_assembly```

**Off by 1 error at very start of + strand transcript sequence**

This appears ok with - strand, but on + strand if you pick the 1st base of transcript, eg "NM_198689.2:n.1C>G" it appears the sequence only start rendering 1 base after where it should, eg [NM_198689.2](https://www.ncbi.nlm.nih.gov/nuccore/NM_198689.2?report=fasta) starts with ```CATCTCCTCCAG``` (so missing "C") below:

```
In [79]: print(pp.display(var_c))
hgvs_g    : NC_000021.9:g.44600597C>G
hgvs_c    : NM_198689.2(KRTAP10-7):c.-25C>G
hgvs_p    : NP_941962.1:p.?
          :    44,600,590          44,600,610          44,600,630
chrom pos :    |    .    |    .    |    .    |    .    |    .    |    .  
seq    -> : CACTCACTCCCATCTCCTCCAGTTCAATCCCCAGCATGGCTGCGTCCACTATGTCTGTCTG
region    :           G                                                  
tx seq -> :            atctcctccagttcaatccccagcATGGCTGCGTCCACTATGTCTGTCTG
tx pos    :                |    .    |    .    |   .    |    .    |    . 
          :                -20       -10       1        10        20
aa seq -> :                                    MetAlaAlaSerThrMetSerValCy
aa pos    :                                                ...           
          :                                    1                        
```

**Exception with intergenic g.HGVS**

```
hgvs/pretty/datacompiler.py:354, in DataCompiler.data(self, var_g, var_c_or_n, display_start, display_end)
    351     pd = reversed(position_details)
    352     position_details = list(pd)
--> 354 is_rna = var_c_or_n.ac.startswith("NR_") # not sure how to check this for ENSTs

AttributeError: 'NoneType' object has no attribute 'ac'
```

fix would be to replace it with ```is_rna = var_c_or_n and var_c_or_n.ac.startswith("NR_") ```

On the comment ```not sure how to check this for ENSTs``` - perhaps it's better to just store whether the type is 'n' or 'c'? That's what ultimately you care about in ```hgvs.pretty.console.tx_pos.TxRulerRenderer.display```

```
            c_pos = pdata.c_pos
            interval = pdata.c_interval
            if is_rna:
                c_pos = pdata.n_pos
                interval = pdata.n_interval
```

**Comment by andreasprlic** [comment]
Thanks, these are great comments. Thank you for looking into the details of this. Luckily this was easy to address. PrettyPrint now only has one assembly mapper, and I fixed the problem with the first base on the transcript, as well as with intergenic variants.

**Comment by andreasprlic** [comment]
@reece any more requests for changes from your side?

File: misc/docker-compose.yml [hunk 1]
```diff
@@ -33,7 +33,9 @@ services:
     volumes:
       - uta_vol:/var/lib/postgresql/data
     network_mode: host
-
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
```
reece, line 37: Why is this needed? You should be able to log into all instances with anonymous:anonymous.

src/hgvs/easy.py
```diff
@@ -66,6 +67,8 @@
 t_to_g = projector.t_to_g
 t_to_p = projector.t_to_p
 get_relevant_transcripts = am38.relevant_transcripts
+pretty37 = PrettyPrint(hdp, am37, useColor=True, showLegend=True)
```
reece, line 70: Please use snake case for consistency with rest of code base.

src/hgvs/pretty/console/chrom_seq_renderer.py
reece, line 1: Please make the the class name and module name correspond as with the rest of the code base. E.g., alignmentmapper.py contains the AlignmentMapper class)

src/hgvs/pretty/console/chrom_seq_renderer.py
```diff
@@ -0,0 +1,43 @@
...
+
+        var_seq = ""
+        for p in data.position_details:
+            c = p.ref
+
+            if not c:
+                var_seq += "."
+                continue
+
+            if self.config.useColor:
```
reece, line 28: Please consider putting a colormap in BasicRenderer to ensure color consistency in both directions.

Or maybe in constants.py is even better.

reece (changes requested):
Such awesome work, @andreasprlic!

Some minor requests and comments here, but nothing that should be a show stopper.

misc/docker-compose.yml
andreasprlic: removed it.

src/hgvs/easy.py
andreasprlic: done

src/hgvs/pretty/console/chrom_seq_renderer.py
andreasprlic: done

src/hgvs/pretty/console/chrom_seq_renderer.py
andreasprlic: done

andreasprlic:
comments have been addressed

reece (approved):
Fabulous work, @andreasprlic . Thank you for this excellent contribution!

PR #748, opened 2024-09-15
Fix variant region for ins and dup on intron-exon boundary

### LINKED ISSUE #655: Inconsistencies across intron/exon boundaries

b0d0nne11:
Fixes #655 

This PR is offered as an alternative to https://github.com/biocommons/hgvs/pull/709. It solves the same problem, that certain ins or dup variants that occur at the intron/exon boundaries have non-zero offsets but are still coding. The approach taken here varies from the earlier PR in a few ways:
- It avoids overriding any methods
- It keeps most of the replace reference changes to the variant mapper
- It avoids the intermediate mapping operations
- It keeps the new tests to a single new issues test file


b0d0nne11:
@reece and @andreasprlic would you mind taking a look at this and letting me know if you like the approach here better?

andreasprlic:
Thanks, this looks technically more concise to me. I have a scientific question though: Is it established that insertions at the intron/exon boundary always end up as part of the exon? 

In other words, do we always want to report such insertions as frameshift variants, or is there a chance they are actually not impacting the coding sequence and remain part of the intron? If we are not sure, it feels as if there should be some config, that would allow to customize the behavior of the code. I am thinking of something like the `cross_boundaries` parameter in normalization, where we can influence, if we want to normalize across intron / exon boundaries.

b0d0nne11:
AFAIK they will always be coding, along with the new conditions for duplications that were added. I can certainly create a feature flag for this but IMO I don't think its needed. I think the only thing that would change would be that these niche mapping cases would again return incorrect results. @gostachowiak do you have anything you want to add?

gostachowiak:
@andreasprlic  For the scientific question-- no, it is not established that insertions at the boundary will always be part of the exon.  For any given mutation, it is impossible to know what will happen in real life.  

The idea here is that since the entire intron (including canonical splice sites) is intact, it will default to treating the inserted material as being part of the exon.  Since the canonical splice site is well-defined, this seems like the most rational default assumption.

I believe this is strictly better than the current behavior, where sometimes insertions at boundaries are treated as intronic, and sometimes as exonic, based on what offset values happen to appear in the nomenclature.  The updated behavior is also consistent with other tools such as Mutalyzer. 

A feature flag would make sense, although in my opinion it would be better to deal with in a follow-up issue, because (1) it would be non-trivial, and (2) the current pull request at least establishes a consistent baseline (i.e. eliminates the arbitrary/inconsistent results that are currently returned).

andreasprlic:
Thanks @gostachowiak for the explanation. I agree, more consistent behavior would be desirable. 

Having read your comment, my first thought was "this will make the variants look more pathogenic as a default". As such I had a chat with a variant scientist about this issue. The recommendation from that person is that such variants always need to be reviewed carefully and that as a general rule the goal would be to try to represent in the more benign representation, which would be as an intronic variant and potentially impacting splicing. (3' shuffling rule in HGVS nomenclature is not helpful here and would not play a role in the naming)

As such, how hard would it be to have two options here: move such insertions into the intron and as an alternative move into exon? (One could create synonyms and let a scientist chose the preferred name) 

From a "hgvs as a library" perspective I can easily see that some labs have certain SOPs how they would report out variants in certain situations, such as this. However it is better if the library offers choice, rather than impose behavior when there are more than one possible answers.

b0d0nne11:
Sorry I missed this for so long. Yes, I can make this change. It doesn't look too hard so should have something here soon. If you have any recommendations for the configuration value name that would be great. I'll just put a placeholder in there for now.

@jsstevenson and @andreasprlic

b0d0nne11:
Added `hgvs.global_config.mapping.ref_at_boundary_is_intronic` which defaults to `True` which I believe is the original behavior. I'm not sure is this name makes sense. I will happily change it if needed. I also looked into adding it as an optional parameter on `VariantMapper` and `AssemblyMapper` but it would have to be passed way down through the callstack to where it actually gets used in `AltSeqBuilder`. Its possible but it would be pretty ugly.

b0d0nne11:
Also just to note, the automated tests appear to be broken which I don't have time to look into today but the tests do pass locally for me fwiw.

b0d0nne11:
@gostachowiak just pointed out to me that my change is not quite correct. As written, if `ref_at_boundary_is_intronic` is set to false the original logic is maintained but that doesn't mean the all refs at the boundary are labeled as intronic. They could be labeled as intronic or exonic depending on which side of the boundary they fall on. In order to make it consistent I need to add logic and test cases to handle the other side of the boundary (the side containing the exon). I'm going to try to accommodate this but it will mean that some variants will be mapped differently regardless of how things are configured.

src/hgvs/variantmapper.py
```diff
@@ -491,12 +519,15 @@ def _replace_reference(self, var):
...
             return var
 
-        seq = self.hdp.get_seq(var.ac, seq_start, seq_end)
+        seq = self.hdp.get_seq(ac, seq_start, seq_end)
 
         if len(seq) != seq_end - seq_start:
             # tried to read beyond seq end; this is an out-of-bounds variant
             return var
 
+        if _type == "g" and mapper and mapper.strand == -1:
```
andreasprlic, line 528: Interesting, just out of curiosity, when can the strand be negative for a g.? Inversions?

tests/issues/test_655.py
```diff
@@ -0,0 +1,155 @@
...
+
+@pytest.fixture(scope="module")
+def mock_vm(mock_hdp):
+    return hgvs.variantmapper.VariantMapper(mock_hdp, prevalidation_level="INTRINSIC")
+
+
+@pytest.fixture()
+def ref_at_boundary_is_exonic():
+    original = hgvs.global_config.mapping.ref_at_boundary_is_intronic
+    hgvs.global_config.mapping.ref_at_boundary_is_intronic = False
```
andreasprlic, line 130: If I am reading this correctly, we are running all tests with this fixture as `ref_at_boundary_is_intronic = False`, but the default is `True`. Do we have tests that verify the default behavior as well? Besides that, this all looks good to me. Thank you for the detailed oriented work that is necessary for this PR!

tests/data/gcp/real.tsv
```diff
@@ -50,7 +50,7 @@ ID00048	NC_000010.10:g.89692922T>C	NM_000314.4:c.406T>C	NP_000305.3:p.(Cys136Arg
 ID00049	NC_000010.10:g.89692921dupA	NM_000314.4:c.405dupA	NP_000305.3:p.(Cys136Metfs*44)
 ID00050	NC_000010.10:g.89692923_89692939delGTGCATATTTATTACAT	NM_000314.4:c.407_423delGTGCATATTTATTACAT	NP_000305.3:p.(Cys136Serfs*38)
 ID00051	NC_000010.10:g.89712015C>A	NM_000314.4:c.633C>A	NP_000305.3:p.(Cys211*)
-ID00052	NC_000010.10:g.89685314dupT	NM_000314.4:c.209dupT	NP_000305.3:p.(Cys71Leufs*3)
```
gostachowiak, line 53: I verified that this is indeed a duplication of the last base of the exon.  Therefore it will now be p.? since the requested default behavior was to have insertions at the exon-intron boundary count as intronic.

b0d0nne11:
Made a few updates:
- Rebased on main
- Added cases for duplications that are adjacent to the intron-exon boundary but are in the exon. This ensures that all insertions on the boundary or duplications adjacent to the boundary are treated consistently.
- Renamed the new configuration variable `global_config.mapping.ins_at_boundary_is_intronic`

src/hgvs/variantmapper.py
b0d0nne11: I'm not really sure. I don't have a simple, science-y explanation for why this is the case. My fist pass at this PR didn't include it and we noticed some mappings gave the reverse compliments of the expected values. After some more digging we found that only g type variants on the negative strand as described by the alignment mapper were affected.

tests/issues/test_655.py
b0d0nne11: In my most recent changes I updated the tests I added in this PR to cover both cases of the new configuration value. In addition, all of the existing tests also cover the case where `ins_at_boundary_is_intronic` is `True` since that's the default value.

b0d0nne11:
@andreasprlic @jsstevenson can you take the stale tag off of this?

b0d0nne11:
Rebased on main and updated new tests to use `conftest.py` fixtures.

b0d0nne11:
@reece et al., I'm not sure what to make of these latest test failures. They seem to indicate that the local seqrepo service isn't available on `localhost:5000`. This doesn't make much sense to me. Why aren't other tests that use conftest fixtures also failing? Is there an override for something that I'm missing? FWIW, `make test` passes for me locally when I have a seqrepo service running.

andreasprlic:
I had the same test issue for one of my PRs. I think merging back the main branch into my feature branch fixed this. Can you try that?

b0d0nne11:
Hi @andreasprlic, I believe the branch is already on the latest main. Is it possible for you all to re-run the tests? Otherwise I can make a whitespace-only change somewhere.

b0d0nne11:
I think I figured it out. The `vm`, `am37`, and `am38` fixtures in `conftest.py` need to depend on the `hdp` fixture. Otherwise those fixtures will refer to the globals in `hgvs.easy` using a default UTA connector rather than the UTA instance shimmed in on lines 24-26 of `conftest.py`.

b0d0nne11:
Hi @reece @andreasprlic et al., have you had a chance to look at this recently? Just wondering if you have any other thoughts or edits? Thanks!

andreasprlic:
What is the c. for `NC_000004.12:g.186276230_186276231insTCC` ?

b0d0nne11:
With `ref_at_boundary_is_intronic = True`:
```
In [1]: import hgvs.config

In [2]: hgvs.config.global_config.mapping.ins_at_boundary_is_intronic = True

In [3]: global_config.mapping.ins_at_boundary_is_intronic
Out[3]: True

In [4]: var_g = parse("NC_000004.12:g.186276230_186276231insTCC")

In [5]: var_c = am38.g_to_c(var_g, 'NM_000128.4')

In [6]: str(var_c)
Out[6]: 'NM_000128.4:c.596-1_596insTCC'

In [7]: var_p = am38.c_to_p(var_c)

In [8]: str(var_p)
Out[8]: 'NP_000119.1:p.?'

In [9]: var_g = parse("NC_000002.12:g.151501482_151501484dup")

In [10]: var_c = am38.g_to_c(var_g, 'NM_001271208.2')

In [11]: str(var_c)
Out[11]: 'NM_001271208.2:c.24034-1_24035dup'

In [12]: var_p = am38.c_to_p(var_c)

In [13]: str(var_p)
Out[13]: 'NP_001258137.2:p.?'
```
And with `ref_at_boundary_is_intronic = False`:
```
In [1]: import hgvs.config

In [2]: hgvs.config.global_config.mapping.ins_at_boundary_is_intronic = False

In [3]: global_config.mapping.ins_at_boundary_is_intronic
Out[3]: False

In [4]: var_g = parse("NC_000004.12:g.186276230_186276231insTCC")

In [5]: var_c = am38.g_to_c(var_g, 'NM_000128.4')

In [6]: str(var_c)
Out[6]: 'NM_000128.4:c.596-1_596insTCC'

In [7]: var_p = am38.c_to_p(var_c)

In [8]: str(var_p)
Out[8]: 'NP_000119.1:p.(Ala199delinsValPro)'

In [9]: var_g = parse("NC_000002.12:g.151501482_151501484dup")

In [10]: var_c = am38.g_to_c(var_g, 'NM_001271208.2')

In [11]: str(var_c)
Out[11]: 'NM_001271208.2:c.24034-1_24035dup'

In [12]: var_p = am38.c_to_p(var_c)

In [13]: str(var_p)
Out[13]: 'NP_001258137.2:p.?'
```

andreasprlic:
Thanks this all looks good to me now! Thanks for being persistent and carrying this so far.

andreasprlic:
And now i have an interesting permission issue and can't approve this MR .

b0d0nne11:
Awesome! I think I need to rebase it on master before you can merge. I'll do that today.

b0d0nne11:
@andreasprlic that should do it. :)

andreasprlic:
@reece I can't approve this PR somehow. Any idea why? I have permissions to apply the admin bypass and force a merge without approval, but can't approve. I see the message `Code owner review required by reviewers with write access.`

theferrit32:
@andreasprlic try now

andreasprlic (approved):
Thank you for all your effort and being persistent on this one. This looks good to me now.

PR #755, opened 2025-01-03
fix(test): add vcr_config for cassette management and re-run cassettes

korikuzma:
* set record_mode to fix failing tests
  * ci uses 'none'
  * default uses 'new_episodes' (can overide via VCR_RECORD_MODE env var)
* re-ran cassettes to update with the correct config


jsstevenson:
shouldn't vcr mode be `none` in CI?

tests/conftest.py
```diff
@@ -1,11 +1,23 @@
...
 
 import pytest
 from support import CACHE
 
 import hgvs.easy
 from hgvs.extras.babelfish import Babelfish
 
+
+@pytest.fixture(scope="function")
+def vcr_config(request):
```
jsstevenson, line 12:
```suggestion
def vcr_config(request):
    """See https://pytest-vcr.readthedocs.io/en/latest/configuration/"""
```

jsstevenson (approved):

jsstevenson (approved):

Issue #551, opened 2019-02-24
Clean-up testing


reece:
The hgvs package has many tests, including many tests for essentially the same code paths. For this issue

* Make per-gene tests optional. Instead, test by functionality.
* Fail fast. Write an early smoke test of basic functionality. Failures in this smoke test should stop testing immediately.
* Use fixtures.
* Migrate legacy unittest tests.

PR #756, opened 2025-02-13
Update test cache

reece:
Updating the test cache turned out to be a total mess. For an explanation of the issues, see #760 and #761 (and #551 for general testing cleanup work).

In the meantime, this PR implements a simple workaround to build the test cache iteratively to avoid the last-write-wins problem so that testing works again.


jsstevenson (approved):
👍 seems fine for now. Thanks for putting the time in on this one!

Issue #764, opened 2025-03-04
Fix g to c mappings where the end position is after the coding region


b0d0nne11:
The variant mapper `g_to_c` function throws an exception when trying to map insertion variants where the end position is between the coding region and UTR. This is due to some position arithmetic being attempted on positions with different datums.

```
In [1]: var_g = parse("NC_000007.13:g.106545832_106545833insAG")

In [2]: var_c = g_to_c(var_g, 'NM_002649.2')
---------------------------------------------------------------------------
HGVSUnsupportedOperationError             Traceback (most recent call last)
Cell In[2], line 1
----> 1 var_c = g_to_c(var_g, 'NM_002649.2')

File ~/workspace/go/hgvs/src/hgvs/assemblymapper.py:109, in AssemblyMapper.g_to_c(self, var_g, tx_ac)
    108 def g_to_c(self, var_g, tx_ac):
--> 109     var_out = super(AssemblyMapper, self).g_to_c(
    110         var_g, tx_ac, alt_aln_method=self.alt_aln_method
    111     )
    112     return self._maybe_normalize(var_out)

File ~/workspace/go/hgvs/src/hgvs/variantmapper.py:271, in VariantMapper.g_to_c(self, var_g, tx_ac, alt_aln_method)
    265 if not pos_c.uncertain:
    266     edit_c = self._convert_edit_check_strand(mapper.strand, var_g.posedit.edit)
    267     if (
    268         edit_c.type == "ins"
    269         and pos_c.start.offset == 0
    270         and pos_c.end.offset == 0
--> 271         and pos_c.end - pos_c.start > 1
    272     ):
    273         pos_c.start.base += 1
    274         pos_c.end.base -= 1

File ~/workspace/go/hgvs/src/hgvs/location.py:178, in BaseOffsetPosition.__sub__(lhs, rhs)
    176 assert type(lhs) == type(rhs), "Cannot substract coordinates of different representations"
    177 if lhs.datum != rhs.datum:
--> 178     raise HGVSUnsupportedOperationError(
    179         "Interval length measured from different datums is ill-defined"
    180     )
    181 if lhs.base == rhs.base:
    182     return lhs.offset - rhs.offset

HGVSUnsupportedOperationError: Interval length measured from different datums is ill-defined
```

I believe this should map to `NM_002649.2:c.3309_*1insAG`.

Issue #761, opened 2025-02-27
Create only one UTA db connection during testing


reece:
**Is your feature request related to a problem? Please describe.**

The root cause of #760 is that we open multiple database connections, each of which has its own cache. When the tests shut down, only one cache survives, and other cached data is lost. Relearning on the incomplete cache builds up additional cached data, which is why the `make test-relearn-iterative` ~hack~ temporary fix in #756 works.

**Describe the solution you'd like**

The root cause is that we are using a cache in which the last write wins, and using that in the context of a bunch of unnecessary connections. For this issue, eliminate redundant UTA connections during testing. (A separate issue should deal with updating caching; see #450.) 

**Describe alternatives you've considered**

**Additional context**

Issue #760, opened 2025-02-26
Test data caching is incomplete


reece:
**Describe the bug**

Learning caching appears to be unreliable: A test run in learn mode appears to succeed, but then fails for some tests immediately after. Learning again (on top of existing data) sometimes fixes the error. 

```
# remove the cache
» \rm tests/data/cache-py3.hdp

# test and learn on tests/issues
» HGVS_CACHE_MODE=learn pytest --no-cov -x tests/issues/
====================================== test session starts =======================================
platform linux -- Python 3.12.3, pytest-8.3.4, pluggy-1.5.0
HGVS_CACHE_MODE=learn
HGVS_SEQREPO_URL=http://localhost:5000/seqrepo
UTA_DB_URL=postgresql://anonymous@localhost:5432/uta/uta_20241220
hgvs.easy.hdp.url=postgresql://anonymous@localhost:5432/uta/uta_20241220
hgvs.easy.hdp.seqfetcher.source='SeqRepo REST (http://localhost:5000/seqrepo)'
rootdir: /home/reece/projects/biocommons/hgvs
⋮
================================= 52 passed, 3 warnings in 0.97s =================================

# now try to test with existing data (no fallback)
» HGVS_CACHE_MODE=run pytest --no-cov -x tests/issues/  
====================================== test session starts =======================================
platform linux -- Python 3.12.3, pytest-8.3.4, pluggy-1.5.0
⋮
collected 52 items                                                                               

tests/issues/test_02xx.py .F

============================================ FAILURES ============================================
_______________________ Test_Issues.test_285_partial_palindrome_inversion ________________________

self = <test_02xx.Test_Issues testMethod=test_285_partial_palindrome_inversion>

    def test_285_partial_palindrome_inversion(self):
        # https://github.com/biocommons/hgvs/issues/285/
        # Inversion mapping is not palindrome aware
        v = self.hp.parse_hgvs_variant("NM_000088.3:c.589_600inv")
>       vn = self.vn.normalize(v)

tests/issues/test_02xx.py:47: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/hgvs/normalizer.py:98: in normalize
    var = self.vm.c_to_n(var)
src/hgvs/variantmapper.py:353: in c_to_n
    self._validator.validate(var_c)
[...]

PR #765, opened 2025-03-04
Fix exception in g_to_c mapping

b0d0nne11:
Fixes https://github.com/biocommons/hgvs/issues/764.


src/hgvs/variantmapper.py
```diff
@@ -268,6 +268,7 @@ def g_to_c(self, var_g, tx_ac, alt_aln_method=hgvs.global_config.mapping.alt_aln
                 edit_c.type == "ins"
                 and pos_c.start.offset == 0
                 and pos_c.end.offset == 0
+                and pos_c.start.datum == pos_c.end.datum
```
reece, line 271: Great catch!

reece (changes requested):
Thanks Brendan! This looks good with some very small change requests.

conftest.py has most of the fixtures that you define in test_764.py. Would you please use those instead? I think it should be sufficient to remove the `hdp` and `am37` fixtures, and renaming `hp` to `parser`.

You'll need to define your own `vm` for now. Although, for bonus points, you could move it to conftest.

You might very well have modeled this on existing tests, but I've recently uncovered challenges with the current layout. See #760 and #761 for details.

The path forward is to consolidate fixtures in conftest.py.

Thanks!

b0d0nne11:
Thanks @reece. Using the `conftest.py` fixtures was much easier.

reece (approved):

Issue #663, opened 2023-07-05
mitochondria (m.) assembly mapping


davmlaw:
RefSeq does not have MT transcripts, but Ensembl does, eg [ENST00000361381.2](https://grch37.ensembl.org/Homo_sapiens/Transcript/Summary?g=ENSG00000198886;r=MT:10760-12137;t=ENST00000361381)

There is currently no "m" (ie m_to_c) etc methods in AssemblyMapper, so you can't convert these easily.

Q. Should we add m_to_x and x_to_m methods in AssemblyMapper?

```
var_m = parser.parse("NC_012920.1:m.11038delA")
am.g_to_c(var_m, "ENST00000361381.2")
File /usr/local/lib/python3.10/dist-packages/hgvs/variantmapper.py:235, in VariantMapper.g_to_c(self, var_g, tx_ac, alt_aln_method)
    222 """Given a parsed g. variant, return a c. variant on the specified
    223 transcript using the specified alignment method (default is
    224 "splign" from NCBI).
   (...)
    231 
    232 """
    234 if not (var_g.type == "g"):
--> 235     raise HGVSInvalidVariantError("Expected a g. variant; got " + str(var_g))
    236 if self._validator:
    237     self._validator.validate(var_g)

HGVSInvalidVariantError: Expected a g. variant; got NC_012920.1:m.11038del
```

It's pretty trivial to get it to work, by switching the 'm' to a 'g' (this is probably how you'd implement it)

```
var_g = deepcopy(var_m)
var_g.type = 'g'
In [19]: print(am.g_to_c(var_g, "ENST00000361381.2"))
ENST00000361381.2:c.279del
```

I don't think you can reproduce this using UTA as it doesn't have Ensembl transcripts, but here's the cdot code:

```
import hgvs.assemblymapper

from copy import deepcopy
from hgvs.parser import Parser
from cdot.hgvs.dataproviders import JSONDataProvider, RESTDataProvider, FastaSeqFetcher

seqfetcher = FastaSeqFetcher("/data/annotation/fasta/GCF_000001405.39_GRCh38.p13_genomic.fna.gz")
hdp = JSONDataProvider(["/data/annotation/cdot/ensembl/GRCh38/cdot-0.2.17.ensembl.grch38.json.gz"], seqfetcher=seqfetcher)
am = hgvs.assemblymapper.AssemblyMapper(hdp, assembly_name='GRCh38', alt_aln_method='splign', replace_reference=True)

parser = Parser()

[...]

PR #778, opened 2025-07-15
Add support for mitochondria (m.) assembly mapping

b0d0nne11:
Fixes https://github.com/biocommons/hgvs/issues/663.

Changes:
- Allow m. variants in `g_to_t`, `g_to_c`, and `g_to_n` mapper functions.
- Added a `translation_table` parameter to the `c_to_p` mapper functions that defaults to `TranslationTable.standard`. This allows us to pass `TranslationTable.vertebrate_mitochondrial` when mapping m. variants. The downside of this approach is that you have to know (1) that the c. came from a m. rather than a g. and (2) that you need to use a different translation table. This would also allow you to pass other translation tables like `TranslationTable.selenocysteine` but I'm not sure when that would come up.
- Fix transcript sequence lengths when the mitochondrial stop codon `TAA` is abbreviated. This was required to handle the transcript sequences that weren't divisible by 3.
- Added tests based on the variants in [chrMT_test_variants.csv](https://github.com/biocommons/hackathon-2023/files/12446748/chrMT_test_variants.csv) I got from https://github.com/biocommons/hackathon/issues/4.


reece:
Thanks Brendan! I'll look at this PR this weekend.

src/hgvs/utils/reftranscriptdata.py
```diff
@@ -20,14 +20,21 @@ def __init__(self, hdp, tx_ac, pro_ac):
 
         # coding sequences that are not divisable by 3 are not yet supported
         tx_seq_to_translate = tx_seq[cds_start - 1 : cds_stop]
+        if translation_table == TranslationTable.vertebrate_mitochondrial:
+            if len(tx_seq_to_translate) % 3 == 1 and tx_seq_to_translate[-1] == "T":
```
reece, line 24: @b0d0nne11 : I'm not familiar with this mito translation rule. It appears to me that you're effectively creating an explicit stop codon if it's incomplete. (TAA = UAA is one of 4 valid mito stop codons)

Is that right?

reece (changes requested):
Hi @b0d0nne11 . This looks great. Thank you. 

I have only one question/request: Can you please add a simple comment in reftranscriptdata.py to explain what that code's doing? Otherwise, terrific. Thanks!

PR #794, opened 2025-09-05
225 uncertain ranges

davmlaw:
Fix broken Babelfish VCF. Tests pass (on my machine, we'll see on actions)


andreasprlic:
Should there be a new unit test for this?

PR #799, opened 2025-09-08
migrate to new python packaging

reece:
This PR migrates hgvs to the new python packaging.

To test:
- `rm -fr venv .venv`
- `make devready`
- `make test`


andreasprlic (dismissed):
LGTM

tests/conftest.py
```diff
@@ -20,11 +20,11 @@ def remove_response_headers(response):
 
 
 @pytest.fixture(scope="module")
-def vcr_config(request):
+def vcr_config(request):  # noqa: ARG001
     """See https://pytest-vcr.readthedocs.io/en/latest/configuration/"""
     return {
         "match_on": ["method", "uri"],
-        "record_mode": os.environ.get("VCR_RECORD_MODE", "new_episodes"),
+        "record_mode": os.environ.get("VCR_RECORD_MODE", "none"),
```
korikuzma, line 27: I'm wondering if in CI we should set the mode to none and have new_episodes be the default for dev?

LICENSE.txt
```diff
@@ -179,15 +178,15 @@
...
+      boilerplate notice, with the fields enclosed by brackets "{}"
       replaced with your own identifying information. (Don't include
       the brackets!)  The text should be enclosed in the appropriate
       comment syntax for the file format. We also recommend that a
       file or class name and description of purpose be included on the
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2019 bioutils Contributors
```
korikuzma: Should this be hgvs?
```suggestion
   Copyright 2019 hgvs Contributors
```

Makefile
```diff
@@ -5,199 +5,114 @@
...
-DOC_TESTS:=docs hgvs ./README.rst
-
 # TESTING sources
 # UTA_DB_URL must be accessible at test time (for now)
-export HGVS_CACHE_MODE=
+# export HGVS_CACHE_MODE=
 _UTAPW=anonymous
 export UTA_DB_URL=postgresql://anonymous:${_UTAPW}@uta.biocommons.org:5432/uta/uta_20241220
-export HGVS_SEQREPO_URL=http://localhost:5000/seqrepo
+# export HGVS_SEQREPO_URL=http://localhost:5000/seqrepo
```
korikuzma: I think we should remove commented out code (or if there's a reason why we should leave it, maybe add an additional comment?)

korikuzma (commented):
Minor comments

tests/conftest.py
reece: I'd prefer to keep this change. The rationale is that most changes probably won't need new testing data, and it's good to make developers aware of code that does require new testing data.

LICENSE.txt
reece: Good catch!

Issue #710, opened 2023-11-10
How to properly handle selenoproteins


holtgrewe:
One example is gene `SELENON` with transcript `NM_206926.2`. Here, an alternative translation table must:

`Note=UGA stop codon recoded as selenocysteine`

Example line from the NCBI GFF3 file:

```
NC_000001.10    BestRefSeq      CDS     26126722        26126904        .       +       0       ID=cds-NP_996809.1;Parent=rna-NM_206926.2;Dbxref=CCDS:CCDS41283.1,GeneID:57190,Genbank:NP_996809.1,HGNC:HGNC:15999,MIM:606210;Name=NP_996809.1;Note=UGA stop codon recoded as selenocysteine%3B isoform 1 is encoded by transcript variant 1;gbkey=CDS;gene=SELENON;product=selenoprotein N isoform 1;protein_id=NP_996809.1;transl_except=(pos:26139280..26139282%2Caa:Sec)
```

biocommons.bioutils already contains the proper table. Is this implemented in UTA/hgvs anywhere? There are some matches of "seleno" in the UTA repository limited to `misc/EnsemblUTA/*`.

holtgrewe:
Ping

andreasprlic:
Fixed by #814

Issue #420, opened 2017-04-04
Selenocysteine not correctly handled by c_to_p


reece:
@veenaraj reported that NC_000001.10:g.26139280T>G is projected by VEP to ENSP00000363434:p.Sec428Gly. hgvs gives NP_065184.2:p.?.

N.B.: c.1384 should map to p.462. I've not confirmed the VEP output. 

Regardless of whether there's a legit bug here, we should make sure that we have tests for TGA → selenocystein/Sec/U and TAG → pyrolysine/Pyl/O.

Invitae internal: https://invitae.jira.com/browse/VDB-565

reece:
hgvs 1.0.0rc1:

```
In [1]: var_g = hp.parse_hgvs_variant("NC_000001.10:g.26139280T>G")

In [2]: var_c = am37.g_to_c(var_g, "NM_020451.2")
/usr/local/share/seqrepo/master/sequences/2016/0824/050233/1472014953.2945154.fa.bgz

In [3]: var_p = am37.c_to_p(var_c)

In [4]: var_c
Out[4]: SequenceVariant(ac=NM_020451.2, type=c, posedit=1384T>G)

In [5]: var_p
Out[5]: SequenceVariant(ac=NP_065184.2, type=p, posedit=None)

In [21]: print(hgvs.utils.context.variant_context_w_alignment(am37, var_c, margin=3))
   NM_020451.2  c       1381 >                                               >     1387 NM_020451.2:c.1384T>G
   NM_020451.2  n       1436 > TGCTGAG                                       >     1442 NM_020451.2:n.1439T>G
                               |||||||                                       7= 
   NC_000001.10 g   26139277 > TGCTGAG                                       > 26139283 NC_000001.10:g.26139280T>G
   NC_000001.10 g   26139277 < ACGACTC                                       < 26139283 NC_000001.10:g.26139280T>G
                                  *                                           NC_000001.10:g.26139280T>G

```

Indeed, this variant is at a TGA codon.

veenaraj:
These are thr esults from VEP 79/ GRch37
Input:
1_26139280_T/G 
Output:
missense_variant  
HGVSc=ENST00000361547.2:c.1384T>G
HGVSp=ENSP00000355141.2:p.Sec462Gly

reece:
Indeed, NP_065184.2 has a U at p.462.

PR #814, opened 2025-12-14
fix for #420  and #710 - selenocysteine support

andreasprlic:
**Summary**:

In this MR we are adding support to create correct hgvs_p names, for variants that impact Selenocysteine positions.

Previously they would yield `p.?`, however now it is possible to create names like `NP_065184.2:p.Sec462Arg`, or `NP_065184.2:p.U127del`

**Details**:

1. Implemented selenocysteine auto-detection in RefTranscriptData.__init__:

- After getting the protein accession, fetches the reference protein sequence

- Checks if it contains 'U' (selenocysteine)

- If found, re-translates using TranslationTable.selenocysteine

- Stores the actual translation table used in self.translation_table

2. Updated VariantMapper.c_to_p to use the detected translation table:

- Passes reference_data.translation_table to AltSeqBuilder so variant sequences use the correct table

Unfortunately also needed to address linting in the files, since I could not commit otherwise.


src/hgvs/variantmapper.py
```diff
@@ -98,18 +96,14 @@ def __init__(
...
         if var_g.type not in "gm":
-            raise HGVSInvalidVariantError(
-                "Expected a g. or m. variant; got " + str(var_g)
-            )
+            raise HGVSInvalidVariantError("Expected a g. or m. variant; got " + str(var_g))
 
         if self._validator:
             self._validator.validate(var_g)
         var_g.fill_ref(self.hdp)
-        mapper = self._fetch_AlignmentMapper(
```
reece, line 112: I suspect that you made this change in order to pacify ruff's N802 warning. 

Throughout biocommons code, names are preserved without modification. I'd like to keep that convention for consistency.

Please approve and merge https://github.com/biocommons/hgvs/pull/815 into main, then merge main into this branch, and then revert the name changes.

reece (changes requested):
Nice work, as usual @andreasprlic. Tests pass for me and the code makes sense.

Please address the minor comment below. Sorry you got tangled up in the rollout of the ruff.

src/hgvs/variantmapper.py
andreasprlic: Done!

andreasprlic:
This PR is not stale. There is something not right with the stale bot. I believe the requested changes were implemented and this would be ready for another round of review.

anju24 (approved):
The requested changes have been addressed. Looks good 👍

reece (approved):
LGTM! Thanks @andreasprlic and @anju24 .

PR #815, opened 2025-12-16
update pyproject.toml from most recent python-package

reece:
Updates pyproject.toml from the python-package template, primarily to ignore ruff warnings that differ from biocommons conventions. These changes should simplify https://github.com/biocommons/hgvs/pull/814/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dev = [
   "sphinxcontrib-fulltoc >= 1.1",
   "toml-sort",
   "tox-uv>=1.28",
-  "vcrpy"
+  "vcrpy",
 ]
 
 [project]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -74,7 +74,7 @@ align = ["uta-align~=0.3"]
 
 [project.urls]
 Documentation = "https://hgvs.readthedocs.io/"
-Homepage = "https://github.com/biocommons/hgvs"
+Homepage = "https://github.com/biocommons/hgvs/"
 Issues = "https://github.com/biocommons/hgvs/issues"
```
korikuzma: Not sure if the slash was added for consistency... if so
```suggestion
Issues = "https://github.com/biocommons/hgvs/issues/"
```

korikuzma (dismissed):

pyproject.toml
reece: Yes, I try to remember to add those. (On some systems, it saves an http redirect delay). Thanks!

PR #823, opened 2026-03-26
Fix/819 Fix incorrect g->c mapping when CIGAR alignment gaps overlap variant interval

### LINKED ISSUE #819: delins adjacent to tx-ref disagreement incorrectly produces a double-gap instead of collapsing to a SNV
**Describe the bug**
When a genomic delins variant spans a position that contains a transcript-reference disagreement (a CIGAR I segment — bases present in the genome but absent from the transcript), the internal mapping creates a double-gap that should collapse to a  simple substitution, but instead produces an incorrect multi-base delins and a  downstream frameshift p-dot.

**Steps to reproduce**

```python
import hgvs.dataproviders.uta as uta
import hgvs.assemblymapper

hdp = uta.connect()
am = hgvs.assemblymapper.AssemblyMapper(hdp, assembly_name="GRCh38")
hp = hgvs.parser.Parser()

var_g = hp.parse("NC_000011.10:g.119027721_119027726delinsTCACA")
var_c = am.g_to_c(var_g, "NM_001164277.1")
print(var_c)   # NM_001467.6:c.527_532delinsTGTGA  ← incorrect
var_p = am.c_to_p(var_c)
print(var_p)   # frameshift ← incorrect
```
Genomic context:
```
         :       119,027,740         119,027,720
chrom pos :  .    |    .    |    .    |    .    |    .    
seq    <- : CGGGATAGACCCCGTGACACCACACCAACAGAGGAAGGAGACAGAG
seq    -> : GCCCTATCTGGGGCACTGTGGTGTGGTTGTCTCCTTCCTCTGTCTC
tx ref dif:                    I                          
region    :                     |----|                    
tx seq -> : GCCCTATCTGGGGCACTGT-GTGTGGTTGTCTCCTTCCTCTGTCTC
tx pos    :   |    .    |    .     |    .    |    .    |  
          :   510       520        530       540       550
aa seq -> : AlaLeuSerGlyAlaLeuC-ysValValValSerPheLeuCysLeu
aa pos    : |||            ...             |||            
          : 170                            180            
```

**Current behavior**

  - g_to_c returns NM_001467.6:c.527_532delinsTGTGA
  - c_to_p returns a frameshift

**Expected behavior**

  The genomic delins only changes a single base relative to the transcript. The tx-ref
  disagreement bases in the flanking I CIGAR segment should be absorbed during mapping,
  collapsing the double-gap to:

  - g_to_c → NM_001164277.1:c.532G>A
  - c_to_p → a missense (not a frameshift)

**Root cause**

  The CIGAR alignment for NM_001164277.1 has an I segment at this position (extra bases
  in the genome not present in the transcript). The current mapping logic does not
  account for adjacent I segments when computing the transcript-coordinate edit, so the
  genome-only bases appear as deleted sequence in the transcript representation,
  producing a spurious multi-base delins instead of a SNV.

andreasprlic:
Closes #819                                                                                                                                                                    
                                                                                                                                                                                 
  **Problem**
                                                                                                                                                                                 
  The g_to_c (and g_to_n) mapping in VariantMapper produced incorrect results for two related cases involving CIGAR alignment gaps:

  1. I-segment strictly inside the variant interval - When a genomic delins spans a position where the CIGAR has an I-segment (a genomic base with no transcript equivalent), both variant endpoints fall in normal = segments, so pos_c.uncertain is False. The previous code path would strand-flip the alt allele without accounting for the missing transcript base, producing an incorrect transcript edit with the wrong length.                                                                                                 
  2. I-segment immediately adjacent to the variant boundary — When a variant endpoint lands exactly at the edge of an I-segment, the previous code would silently ignore the gap, again producing a wrong edit.                                                                                                                                                 
  
  Both cases are manifestations of issue #819 (the "double gap" problem).                                                                                                        
                                                                                                                                   
  **Fix**

  Added helper methods to VariantMapper:

  - `_gap_segments_within_pos_g(mapper, pos_g)` — returns all I/D CIGAR segments strictly inside a genomic interval (replaces the earlier _i_segment_offsets_in_pos_g).            
  - `_variant_has_internal_gap(mapper, var_g)` — returns True when a gap segment lies strictly inside the variant interval (both endpoints in = regions).
  - `_expand_pos_g_for_adjacent_gap(mapper, var_g)` — extends the genomic position to include any I-segment immediately adjacent to the variant boundary.                          
                                                                                                                                   
  The mapping logic in `g_to_c` and `g_to_n` now checks for these cases before the strand-flip path and routes them through _get_altered_tx_sequence(), which correctly filters out  genomic-only bases when reconstructing the transcript edit.                                                                      
                                                                                                                                                                                 
  Tests                                                                                                                            

  Three new test cases in test_hgvs_assemblymapper.py:                                                                                                                           
  
  - NC_000011.10:g.119027721_119027726delinsTCACA → NM_001164277.1:c.532G>A — I-segment adjacent to (but outside) the variant interval                                           
  - NC_000011.10:g.119027726_119027728delinsTT → NM_001164277.1:c.526_527delinsAA — I-segment strictly inside the variant interval 
  - NC_000002.12:g.73385901_73385903del → NM_015120.4:c.34_36del — variant spanning a D-segment boundary


reece:
@andreasprlic : I see that this is PR is ready for review, but also that you added "help wanted" and that there's a conflict with cache-py3.hdp. Should I review or wait?

andreasprlic:
oops I missed the conflict. The help wanted was to indicate that I would like to get this reviewed :-) Will fix the conflict asap.

reece:
I'd take on the conflict, but if I commit I can't review.

reece (approved):

PR #825, opened 2026-04-14
fix(uta): decode URL-encoded creds from UTA_DB_URL

### LINKED ISSUE #824: Handle special characters in DB URL passwords
**Describe the bug**
DB connections fail when passwords contain special characters such as `[ ] @ : /`. After urlparse, the password may remain URL-encoded, causing auth failures.

https://github.com/biocommons/hgvs/blob/8a515cef2df0b04c5feeb347702e9afc4af4c8c6/src/hgvs/dataproviders/uta.py#L703

**To Reproduce**
1. `export UTA_DB_URL=postgresql://uta_user:my%5Bpassword%5D@localhost:5432/uta/uta_20241220`
2. Try connecting to DB
3. Get error:

```
psycopg2.OperationalError: connection to server at "localhost", port 5432 failed: could not initiate GSSAPI security context: Unspecified GSS failure.  Minor code may provide more information: Matching credential not found
connection to server at "localhost", port 5432 failed: FATAL:  password authentication failed for user "uta_user"
```

**Expected behavior**
Connect to DB 

**Additional context**
It should be on user to ensure DB_URL is URL-encoded when containing special characters

korikuzma:
close #824

### DIFF
--- a/src/hgvs/dataproviders/uta.py
+++ b/src/hgvs/dataproviders/uta.py
@@ -66,7 +66,7 @@ def connect(
 ):
     """Connect to a UTA database instance and return a UTA interface instance.
 
-    :param db_url: URL for database connection
+    :param db_url: URL for database connection. Credentials must be URL-encoded if they contain special characters
     :type db_url: string
     :param pooling: whether to use connection pooling (postgresql only)
     :type pooling: bool
@@ -547,7 +547,7 @@ class UTA_postgresql(UTABase):
             port=self.url.port,
             database=self.url.database,
             user=self.url.username,
-            password=self.url.password,
+            password=urlparse.unquote(self.url.password),
             application_name=self.application_name + "/" + hgvs.__version__,
         )
         if self.pooling:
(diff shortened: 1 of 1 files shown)


jsstevenson (dismissed):

theferrit32:
Perhaps we should call unquote on the user name too just in case? Though it's probably less typical for a username to have such characters in it.

src/hgvs/dataproviders/uta.py
```diff
@@ -546,8 +546,8 @@ def _connect(self):
             host=self.url.hostname,
             port=self.url.port,
             database=self.url.database,
-            user=self.url.username,
-            password=self.url.password,
+            user=urlparse.unquot(self.url.username),
```
theferrit32:
```suggestion
            user=urlparse.unquote(self.url.username),
```

theferrit32 (changes requested):

src/hgvs/dataproviders/uta.py
korikuzma: Ugh thanks. My VS Code in browser was being so finicky

theferrit32 (approved):

PR #826, opened 2026-04-15
Example parsley -> pyparsing conversion implementation

### LINKED ISSUE #505: RFC: Develop plan to replace Parsley
We currently use [parsley](http://parsley.readthedocs.io/en/latest/) to parse HGVS expressions. Parsley was last released in 2015. The most recent commit on github is 2017. The project appears to be defunct, which presents a risk for the hgvs package.

Since swapping parser will take significant time and likely create backward-incompatible changes, we should target such a move for a major release (e.g. hgvs 2.x).

https://tomassetti.me/parsing-in-python/ has a great review of Python parser tools.

davmlaw:
Issue #505

Replaces the OMeta/parsley grammar with a pyparsing implementation, rolled out in phases as suggested by @reece: **both grammars ship, OMeta remains the default**, and pyparsing is opt-in.

```python
hgvs.parser.Parser()                            # OMeta/parsley (unchanged default)
hgvs.parser.Parser(grammar_fn="__pyparsing__")  # new pyparsing grammar
hgvs.parser.Parser(grammar_fn="__ometa__")      # explicit; same as the default
hgvs.parser.Parser(grammar_fn="/path/to.pymeta")  # deprecated, warns
```

Suggested phasing:

- **Now (1.x)** — as in this PR: default unchanged, pyparsing available opt-in, `grammar_fn` as a path deprecated.
- **2.0** — switch the default to pyparsing; keep `__ometa__` available.
- **2.1** — remove OMeta/parsley, `sbin/generate_parser.py`, and the generated grammar.

## Validation

The OMeta backend is treated as the reference implementation, and the new grammar was differentially tested against it:

| check | result |
|---|---|
| unique real variants through both backends (gauntlet, clinvar, random-vars, gcp) | **169,575 — zero divergences, zero crashes** |
| `grammar_test.tsv` rule cases under pyparsing | 781/781 pass, agreeing with OMeta on every case |
| grammar rule parity | 120 rules each, no differences either direction |
| public API parity | identical 33 `parse_*` methods |
| parse throughput | pyparsing **2.17x faster** than OMeta |

`tests/test_hgvs_grammar_equivalence.py` keeps this honest in CI: the backends must agree on every gauntlet variant and expose an identical public API.

### One deliberate behaviour change

OMeta silently skips leading spaces/tabs, so `parse_hgvs_variant(" NM_01234.5:c.22+1A>T")` succeeds today. The pyparsing grammar rejects it. HGVS strings contain no whitespace, so the stricter reading is intended — but it is a real change to note at the 2.0 default switch. Both backends reject trailing whitespace and leading newlines. Callers wanting the old leniency should strip their input.

This is the only behavioural difference found anywhere in the corpus above.

### Actively maintained
pyparsing has 20+ years of continuous development. Parsley was last released in 2015, with the most recent commit in 2017. For a clinical-grade package, depending on an abandoned parser is a liability.

### No code generation step (at 2.1, once OMeta is removed)
The grammar becomes ordinary Python in `src/hgvs/grammar.py`, so the offline `sbin/generate_parser.py` step and the ~4,000-line generated `src/hgvs/generated/hgvs_grammar.py` can go, along with the class of "grammar and generated code out of sync" bugs. Both are necessarily retained while OMeta remains the default.

### Better error messages
pyparsing provides [`set_name()`](https://pyparsing-docs.readthedocs.io/en/latest/api/pyparsing/core.html#pyparsing.core.ParserElement.set_name) on rules and detailed `ParseException` objects with position, caret, and context. This makes it straightforward to produce helpful parse errors in a follow-up — for example, `Expected c. position (e.g. 22+1, *10)` instead of a raw PEG expectation list. Parsley's error output was opaque and hard to improve.

### Easier to debug
Each rule is a first-class Python object. You can test, inspect, and step through parse actions in a standard debugger. No OMeta runtime to trace through.

### Faster
2.17x faster than OMeta on real variants, without packrat memoization (see below).

## Notes for reviewers

- **Packrat is deliberately not enabled.** It reads like an obvious win — `hgvs_variant` tries six variant types, `dna_edit` nine — but measured best-of-3 over 5,000 variants it was **1.55x slower** with identical results, so it is off. The 2.17x figure above is without it.
- **pyparsing's whitespace and packrat settings are process-global.** `set_default_whitespace_chars()` also rewrites pyparsing's builtin expressions in place, so an earlier revision of this branch broke any pyparsing grammar built afterwards in the same process — including for users who never opt in. The grammar is now built inside a context manager that sets and restores the default. `tests/test_hgvs_grammar_equivalence.py::TestNoGlobalPyparsingState` guards this.
- **Rules are anchored with an explicit `StringEnd()`** rather than `parse_string(parse_all=True)`, which builds its own `Empty() + StringEnd()` at parse time and would pick up the ambient whitespace default, letting trailing whitespace through.
- **`expose_all_rules=True`** returns 120 methods vs OMeta's 129. The nine missing are OMeta runtime internals (`parse_anything`, `parse_digit`, `parse_token`, ...), not hgvs grammar rules. The default API is unchanged.
- **Dependency footprint** is temporarily larger, not smaller: `parsley` and `pyparsing` are both required while both grammars ship. `parsley` (and `ometa.runtime`) go away at 2.1.

## Provenance

The grammar translation was written by Claude Code. It has not had a line-by-line human review — the confidence above comes from differential testing against the OMeta backend rather than from reading the code, so a human review of `src/hgvs/grammar.py` is still worth doing before the 2.0 default switch. We have an existing spec and a good test corpus, which is what makes this tractable to validate mechanically.


reece:
Looking 👀 .

reece:
This is terrific @davmlaw. Thank you!

What would you think about phasing this rollout like this:

- Now (1.x series): Add back the traditional parsing and preserve default behavior, allow user to choose the pyparsing grammar. Specifically: if `grammar_fn` is (say) `__pyparsing__`, then load use that grammar. If it's specified with another value, warn about deprecation. 

- Later (2.0): Switch default to pyparsing. Consider whether we want to remove ometa or preserve as optional (`__ometa__`?). 

- 2.1: Fully remove ometa

Also, see a minor comment in parser.py.

src/hgvs/parser.py
```diff
@@ -87,15 +107,15 @@ class Parser:
     """
 
     def __init__(self, grammar_fn=None, expose_all_rules=False):
-        bindings = {"hgvs": hgvs, "bioutils": bioutils, "copy": copy}
-        if grammar_fn is None:
-            self._grammar = parsley.wrapGrammar(
-                createParserClass(ometa.runtime.OMetaGrammarBase, bindings)
+        if grammar_fn is not None:
```
reece: Consider using a sentinel to trigger a deprecation warning, like so:

```
_grammar_fn_sentinel = object()    # just a random object

def __init__(self, grammar_fn=_grammar_fn_sentinel):
    if grammar_fn is not _grammar_fn_sentinel:
        warnings.warn("grammar_fn is deprecated", DeprecationWarning, stacklevel=2)
```

This is basically just a way to see whether the user specified `grammar_fn`

davmlaw:
Good idea, running parallel implementations is much safer, we can validate it against the old one.

davmlaw:
@reece thanks for the review — sorry for the slow follow-up. Everything you asked for is in, plus some fixes from validating the new grammar properly. Summary of what changed since your comment:

**Your phased rollout** (32e2a51) — implemented as suggested. Both grammars ship, OMeta stays the default, `grammar_fn="__pyparsing__"` opts in, `grammar_fn="__ometa__"` is explicit, and any other value warns `DeprecationWarning`. I used your sentinel suggestion from the inline comment on `parser.py` verbatim — it's much cleaner than overloading `None`. The PR description now lays out the 1.x → 2.0 → 2.1 phasing.

**Validation** — the OMeta backend is treated as the reference implementation and the new grammar was differentially tested against it:

- **169,575 unique real variants** (gauntlet, clinvar, random-vars, gcp) through both backends: **zero divergences, zero crashes**
- 781/781 `grammar_test.tsv` rule cases pass under pyparsing, agreeing with OMeta on every one
- rule parity 120/120, and an identical public API of 33 `parse_*` methods
- pyparsing is **2.17x faster** than OMeta

**Three fixes that came out of that** (4660a3f):

1. **pyparsing's global state was leaking.** `grammar.py` called `set_default_whitespace_chars("")` and `enable_packrat()` at import, and `parser.py` imports it unconditionally — so merely importing `hgvs.parser` silently broke any pyparsing grammar built afterwards in the same process, *including for users on the OMeta default*. `set_default_whitespace_chars()` also rewrites pyparsing's builtin expressions in place. Worse, the setting was load-bearing but unprotected: elements bake in the whitespace default as they're constructed, so anything resetting the global before `Parser()` was constructed made the grammar accept malformed input like `NM_01234.5 :c.22+1A>T`. The grammar is now built inside a context manager that sets and restores the default.

2. **Packrat is now off.** It seemed like an obvious win given how much the grammar backtracks, but measured best-of-3 over 5,000 variants it was **1.55x slower** with identical results. Dropping it is where most of the 2.17x comes from — the earlier claim in the description was wrong and I've removed it.

3. Rules are anchored with an explicit `StringEnd()` instead of `parse_string(parse_all=True)`, which builds its own `Empty() + StringEnd()` at *parse* time and picks up the ambient whitespace default. This one was sneaky: the old code only rejected trailing whitespace *because of* the global leak in (1), so fixing the leak briefly made `"NM_01234.5:c.22+1A>T "` parse.

**Tests** (f2bbd0f) — the pyparsing grammar had no coverage at all: the completeness test read rule names from the new grammar, but every actual parse test ran through `Parser()`, i.e. OMeta. That's what SonarCloud's "0.0% coverage on new code" was reporting. `test_parser_grammar` now runs against both backends, and there's a new `tests/test_hgvs_grammar_equivalence.py` covering backend agreement on the gauntlet, API parity, and the no-global-leak guarantees. To check the tests actually bite, I mutated the pyparsing grammar (swapped `ref`/`alt` in `dna_subst`): the old suite passed, the new one fails.

**One deliberate behaviour change to flag for 2.0:** OMeta silently skips leading spaces/tabs, so `parse_hgvs_variant(" NM_01234.5:c.22+1A>T")` succeeds today; pyparsing rejects it. HGVS strings contain no whitespace so the stricter reading seems right, and it's the *only* behavioural difference across all 169,575 variants — but it's a real change worth deciding on deliberately rather than inheriting. Both backends reject trailing whitespace and leading newlines. Happy to add a `strip()` if you'd rather preserve the old leniency, though that would loosen the OMeta default too.

**Docs** (64a85bb) — both backends documented in `docs/grammar.rst` and the `Parser` docstring, including the whitespace change. No changelog entry: `docs/changelog/` hasn't been updated since 1.5.3 while releases reached 1.5.6, so I didn't want to invent a process — happy to add one if you'd like.

One caveat worth being explicit about: the grammar translation is Claude Code's work and hasn't had a line-by-line human review. The confidence above comes from differential testing rather than from reading it, so `src/hgvs/grammar.py` still deserves human eyes before the 2.0 default switch. But given the default is unchanged, this should be safe to land on its own terms.

reece (commented):
Amazing work, @davmlaw.

I was thinking that a base class with parsley- and pyparsing-based subclasses. Advantages that I see:

- cleaner structure: moves conditionals from a mult-functional class (Parser) to distinct classes
- better isolation: each backend is totally independent, including not needing the dependencies if not invoked.- better extensibility: if we ever wanted a different backend (TatSu, Lark), we could do that.

Perhaps layout like the following? 
```
src/hgvs/parsers
├── base.py
├── __init__.py
├── parsley_parser.py
└── pyparsing_parser.py
```

As a minor tidiness benefit, we could move the grammar and generated grammar.

See https://gist.github.com/reece/4818230be643b77247d94d6b3e1cd84e for an example.

What say you? I'll take a stab if you don't want to.

reece (approved):

PR #827, opened 2026-04-17
Revert "fix(uta): decode URL-encoded creds from UTA_DB_URL"

reece:
Reverts biocommons/hgvs#825

This change risks breaking existing clients that pass a string that contains special characters.

Data providers should use username and passwords as passed to it; if they need to be unquoted, that should be the caller's responsibility, not the client's. This is akin to unicode encoding/decoding or converting between UTC/local at the code perimeter, not within it.

@korikuzma @andreasprlic : What do you make of this assessment?

### DIFF
--- a/src/hgvs/dataproviders/uta.py
+++ b/src/hgvs/dataproviders/uta.py
@@ -66,7 +66,7 @@ def connect(
 ):
     """Connect to a UTA database instance and return a UTA interface instance.
 
-    :param db_url: URL for database connection. Credentials must be URL-encoded if they contain special characters
+    :param db_url: URL for database connection
     :type db_url: string
     :param pooling: whether to use connection pooling (postgresql only)
     :type pooling: bool
@@ -546,8 +546,8 @@ class UTA_postgresql(UTABase):
             host=self.url.hostname,
             port=self.url.port,
             database=self.url.database,
-            user=urlparse.unquote(self.url.username),
-            password=urlparse.unquote(self.url.password),
+            user=self.url.username,
+            password=self.url.password,
             application_name=self.application_name + "/" + hgvs.__version__,
         )
         if self.pooling:


andreasprlic:
agreed, I found the initial request for this already somewhat interesting, but could not put my finger as nicely on it as you did.

andreasprlic (approved):
LGTM

korikuzma:
@reece @andreasprlic i originally had the DB URL passed directly without URL encoding (we have creds auto generated in our system), but was getting an error in ParseResult. I don’t have the error at the moment (on my phone), but can post Monday. 


This came about when we were trying to deploy a live AnyVar instance for Connect

reece:
reece 0, kori 1

@korikuzma's solution is correct and actually that this has been a latent bug for since the beginning of time.

The db connection string is a URI. Its components (e.g., password) MUST be %-escaped.

Importantly, urllib.parse.urlparse does NOT expand (unescape) during parsing (my misunderstanding). So, I agree with @korikuzma that this needs to be done.

```
>>> from urllib.parse import (urlparse, quote, unquote)
>>> def make_uri(password: str) -> str:
...     return f"postgresql://reece:{quote(password, safe='')}@uta.biocommons.org/"
... 

# simple passwords don't need quoting/unquoting
>>> uri = make_uri("opensesame")
>>> uri
'postgresql://reece:opensesame@uta.biocommons.org/'
>>> urlparse(uri).password
'opensesame'
>>> unquote(urlparse(uri).password)
'opensesame'

# but those with special characters do!
>>> uri = make_uri("but Reece/was wrong!")
>>> uri
'postgresql://reece:but%20Reece%2Fwas%20wrong%21@uta.biocommons.org/'
>>> urlparse(uri).password
'but%20Reece%2Fwas%20wrong%21'
>>> unquote(urlparse(uri).password)
'but Reece/was wrong!'
```

I'm going to close this PR.

PR #830, opened 2026-05-07
Handle ins or dup variants that include a stop codon

### LINKED ISSUE #829: Insertion or duplication that includes a stop codon
This variant is an insertion of 12 bp where the second inserted codon is a stop codon.

```
In [1]: var_c = parse('NM_000159.4:c.1074_1075insAGTTGAAGGACC')

In [2]: var_p = am37.c_to_p(var_c)

In [3]: str(var_p)
Out[3]: 'NP_000150.1:p.(Asp358_Gln359insSerTerArgThr)'
```

We believe this should result in `NP_000150.1:p.(Asp358_Gln359insSerTer)`.

This variant is identical to the above example except that it's missing the last bp.

```
In [4]: var_c = parse('NM_000159.4:c.1074_1075insAGTTGAAGGAC')

In [5]: var_p = am37.c_to_p(var_c)

In [6]: str(var_p)
Out[6]: 'NP_000150.1:p.(Gln359SerfsTer2)'
```

This shouldn't functionally change the result since the difference occurs after the newly added stop codon so the result should still be `NP_000150.1:p.(Asp358_Gln359insSerTer)`.

b0d0nne11:
Fixes https://github.com/biocommons/hgvs/issues/829.


tests/issues/test_829.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+testdata = [
+    ('NM_000159.4:c.1074_1075insAGTTGAAGGACC', 'NP_000150.1:p.(Asp358_Gln359insSerTer)'),
+    ('NM_000159.4:c.1074_1075insAGTTGAAGGAC', 'NP_000150.1:p.(Asp358_Gln359insSerTer)'),
+    ('NM_004380.3:c.138_139insTCATCATGAGCTCCC', 'NP_004371.2:p.(Pro46_Asn47insSerSerTer)'),
+    ('NM_000038.6:c.4189_4197delinsATATAAAAAA', 'NP_000029.2:p.(Glu1397IlefsTer2)'),
+    ('NM_004380.2:c.139delinsTCATCATGAGCTG', 'NP_004371.2:p.(Asn47delinsSerSerTer)'),
+    ('NM_004448.4:c.2339dup', 'NP_004439.2:p.(Tyr781IlefsTer93)')
```
andreasprlic: Can we also have a few tests using one letter code representation?

src/hgvs/utils/altseqbuilder.py
```diff
@@ -319,6 +328,13 @@ def _incorporate_dup(self):
         seq[end:end] = dup_seq
 
         is_frameshift = len(dup_seq) % 3 != 0
+        if end >= cds_start - 1:
+            stop_end = self._insert_stop_seq_end(dup_seq, end, cds_start)
```
andreasprlic, line 332: How does this impact frameshift variants at the end of a transcript, which might result in that there is no stop codon in the extension . I think right now they might get something like ext*?

b0d0nne11:
@andreasprlic I added a few test cases to cover your comments.

b0d0nne11:
@andreasprlic @reece Can you remove the stale label here? Also, do you have any early feedback?

andreasprlic:
I think it looks good. Only question I had was about the frameshift and stop gain variants at the end of a transcript. Can we add a test that includes e.g. `NM_000203.4:c.1960T>C`. That should create a `p.*654Rext*?` pdot. Also `NM_000203.4:c.1730del` -> `fs*?`.

b0d0nne11:
Thanks @andreasprlic. I added those 2 test cases.

andreasprlic:
Thanks, this looks good to me - what's our expectation about SonarCloud warnings? (there is one about a method being too complex). I don't think we have a requirement about fixing those, so I'll go ahead and approve.

andreasprlic (approved):
Overall looks good to me. We should discuss about what our requirements are about SonarCloud analysis warnings / recommendations.

Also, if would be nice to hook up a Claude code review agent.

b0d0nne11:
thanks @andreasprlic! @reece are we gtg here?

PR #831, opened 2026-07-10
fix: "generator didn't stop after raising" errors

igor-petrik-invitae:
As currently implemented, if there is a database issue _while_ inside the `_get_cursor` context, the while will loop and `yield` again. **This is incorrect behavior; a context cannot retry the block it contains.** This causes the manager to raise a `generator didn't stop after raising` exception, masking the original exception and not actually retrying.

The implementation herein retains retries for the initial connection, but moves the `yield` out of the loop. Retrying the block does not work today and was not implemented in this PR.

@andreasprlic @reece


reece (commented):
@igor-petrik-invitae - Good find! This must have been challenging to debug.

My recollection is that the retry logic was added to deal with interrupted TCP/IP connections. The retry logic was intended to wrap the connection itself in order to solve that issue. Your PR retries connections on the initial attempt, but doesn't retry connections that have been dropped, as might happen when there are too many open connections.  @andreasprlic Do you remember any better?

reece (changes requested):
@igor-petrik-invitae - Good find! This must have been challenging to debug.

My recollection is that the retry logic was added to deal with interrupted TCP/IP connections. The retry logic was intended to wrap the connection itself in order to solve that issue. Your PR retries connections on the initial attempt, but doesn't retry connections that have been dropped, as might happen when there are too many open connections. @andreasprlic Do you remember any better?

reece:
@andreasprlic Any comments on this PR?

igor-petrik-invitae:
@reece That is what it looks like the intent was, but it never did or could have done that. My fix is intended to fix the incorrect `GeneratorStop` exception from being raised, without changing the actual behavior. Actually building a system that retries on exceptions within the SQL execution to get the originally intended behavior is a substantially larger rewrite.

PR #870, opened 2026-08-09
fix: isolate network-dependent UTA tests from default test runs

### LINKED ISSUE #867: Isolate network-dependent tests from default runs; document on-demand/learn invocation
## Bug

Several tests silently fall back to the live public UTA instance (`postgresql://anonymous:anonymous@uta.biocommons.org/uta/uta_20210129b`) when a query isn't found in the local test cache (`tests/data/cache-py3.hdp`), instead of failing closed. This lets remote data drift on the public UTA server cause spurious, non-reproducible test failures unrelated to the code change under test.

Observed on `main` (unrelated to a urllib3 dependency bump under investigation — confirmed present both before and after that change):

1. `tests/test_clinvar.py::Test_Clinvar::test_clinvar`
   Fails with:
   ```
   hgvs.exceptions.HGVSInvalidVariantError: insertion length must be 1
     at line 894: BBS4  21739  NC_000015.10:g.72695180dupA NC_000015.9:g.72987521dupA NM_001252678.1:c.-446+8929dupA NM_033028.3:c.28insA NM_033028.4:c.28dupA NR_045565.1:n.77dupA
   ```
   `NM_033028.3:c.28insA` is not valid HGVS (an `ins` of length 1 should be `dup`). Cause is either stale/bad test fixture data sourced from ClinVar, or drift in live UTA transcript/alignment data feeding this test via cache miss → live fallback.

2. `tests/test_hgvs_dataproviders_uta.py::Test_hgvs_dataproviders_uta_with_pooling_without_cache::test_get_gene_info`
   Fails with:
   ```
   assert len(gene_info) == 10
   AssertionError: assert 6 == 10
   ```
   for gene `VHL`. The live `uta_20210129b` instance now returns 6 fields instead of the 10 the test expects — schema/content drift on the remote server.

   Note: this particular test class (`Test_hgvs_dataproviders_uta_with_pooling_without_cache` / its subclass `TestUTAPool`) is not a cache-miss accident — it deliberately calls `hgvs.dataproviders.uta.connect(pooling=True)` with no `mode=`/`cache=` at all, specifically to exercise real connection-pool/connection-loss behavior (`TestUTAPool.test_use_putconn_on_lost_conn`). It always requires network/live-DB access by design.

Both cache-miss-fallback tests log:
```
WARNING  hgvs.dataproviders.uta:uta.py:106 You are executing tests using remote data (postgresql://anonymous:anonymous@uta.biocommons.org/uta/uta_20210129b)
```
confirming they hit the network instead of the bundled cache.

## Impact

- Test runs are non-deterministic across time/environment since they depend on the live public DB's current state.
- Failures get misattributed to whatever change is under review in that session (dependency bumps, refactors, etc.) when the actual cause is unrelated remote data drift.
- No CI signal distinguishes "cache miss, hit live DB" from "cache hit" — the warning is easy to miss in normal test output.
- Default `pytest` runs currently require network connectivity to pass reliably (e.g. `Test_hgvs_dataproviders_uta_with_pooling_without_cache` always dials out), which slows down and destabilizes routine local/CI test runs.

## Goal

The default test run (`pytest`, no extra flags) should be **fully network-isolated** — driven entirely by the bundled `tests/data/cache-py3.hdp` fixture cache and `HGVS_CACHE_MODE=run` (fail-closed on cache miss), with zero live network/DB calls. Tests that inherently require a live UTA connection (pool/connection-loss tests, cache-regeneration) should be excluded by default and run only on demand.

We also need to preserve a documented, easy way to invoke the same tests against the live DB — both to exercise real connection behavior (pooling/failover) and to regenerate/extend `tests/data/cache-py3.hdp` via `HGVS_CACHE_MODE=learn` when new fixture data is needed (i.e. "training" the cache).

## Suggested fix

- Add `@pytest.mark.network` (marker already declared in `pyproject.toml` but currently unused) to tests that always require a live UTA connection, starting with `Test_hgvs_dataproviders_uta_with_pooling_without_cache` (and its subclass `TestUTAPool`).
- Exclude `network`-marked tests from the default `addopts` (alongside the existing `not extra`), so `pytest` alone never touches the network.
- Document/verify the on-demand invocation for both purposes:
  - Running network-only tests deliberately: `pytest -m network`.
  - Regenerating the fixture cache against live UTA: `HGVS_CACHE_MODE=learn pytest ...` (writes new entries to `tests/data/cache-py3.hdp`).
- Separately (can be follow-up work): regenerate/extend `tests/data/cache-py3.hdp` to cover the currently-missing queries, and fix or update the two stale assertions/fixtures above (`test_clinvar.py`, `test_get_gene_info`'s VHL field-count expectation) to match correct/expected data.

---
Related: #849 (parent tracking issue for hdp/testing devops pain)

reece:
## Summary

- Marks `Test_hgvs_dataproviders_uta_with_pooling_without_cache` (and its subclass `TestUTAPool`) with `@pytest.mark.network` — this class connects with no cache and always requires a live UTA database connection by design (it tests connection-pool/failover behavior). The `network` marker was already declared in `pyproject.toml` but unused.
- Excludes `network`-marked tests from default `addopts` (`-m 'not extra and not network'`), so a plain `pytest`/`make test` run never dials out.
- Fixes `tests/conftest.py`, which imported `hgvs.easy` for convenience — but `hgvs.easy` connects with no `mode`/`cache` at *import time* (by design, for library end users), so merely importing it during test collection opened a real network connection to the live UTA DB, before conftest's own cache-backed `connect()` call could take effect. Replaced with direct `hgvs.dataproviders.uta`/`hgvs.parsers.Parser` imports and module-level `_hdp`/`_parser` singletons.
- Updated the `network` marker's description in `pyproject.toml` for clarity: "uncachable tests that always require network connectivity".

**Tests are now fully isolated from network by default.** Verified by running the full suite with an unresolvable `UTA_DB_URL` (simulating no network access) — `make test` passes (420 passed, 44 skipped, 28 deselected, 4 xfailed), where previously it failed with `psycopg.OperationalError: failed to resolve host`. Reece also ran from laptop with network disabled.

Network-only tests can still be run on demand with `pytest -m network` (requires live UTA connectivity). `HGVS_CACHE_MODE=learn`/`verify` continue to work as before for regenerating/verifying the bundled test cache against a live DB.

Fixes #867

## Test plan

- [x] `make test` passes with a deliberately unresolvable `UTA_DB_URL` (no network fallback possible)
- [x] `pytest --collect-only -m network` collects exactly the pool/connection-loss tests
- [x] `pytest -m network` still reaches the live UTA DB when run explicitly
- [x] Full default suite passes unchanged otherwise (420 passed, 44 skipped, 28 deselected, 4 xfailed)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,9 +2,10 @@ import os
 
 import pytest
 
-import hgvs.easy
+import hgvs.dataproviders.uta
 from hgvs.assemblymapper import AssemblyMapper
 from hgvs.extras.babelfish import Babelfish
+from hgvs.parsers import Parser
 from hgvs.variantmapper import VariantMapper
 from support import CACHE
 
@@ -30,14 +31,13 @@ def vcr_config(request):  # noqa: ARG001
     }
 
 
-hgvs.easy.hdp = hgvs.dataproviders.uta.connect(
-    mode=os.environ.get("HGVS_CACHE_MODE", "run"), cache=CACHE
-)
+_hdp = hgvs.dataproviders.uta.connect(mode=os.environ.get("HGVS_CACHE_MODE", "run"), cache=CACHE)
+_parser = Parser()
 
 
 @pytest.fixture(scope="session", autouse=True)
 def parser():
-    return hgvs.easy.parser
+    return _parser
 
 
 @pytest.fixture(scope="session")
@@ -57,7 +57,7 @@ def am38(hdp):
 
 @pytest.fixture(scope="session")
 def hdp():
-    return hgvs.easy.hdp
+    return _hdp
 
 
 @pytest.fixture(scope="session")
@@ -68,8 +68,8 @@ def babelfish38(hdp):
 def pytest_report_header(config):  # noqa: ARG001
     env_vars = ["UTA_DB_URL", "HGVS_SEQREPO_URL", "HGVS_CACHE_MODE"]
     rv = [f"{ev}={os.environ.get(ev)}" for ev in sorted(env_vars)]
-    rv += [f"hgvs.easy.hdp={hgvs.easy.hdp.url}"]
-    rv += [f"{hgvs.easy.hdp.seqfetcher.source=}"]
+    rv += [f"hdp={_hdp.url}"]
+    rv += [f"{_hdp.seqfetcher.source=}"]
     return "\n".join(rv)
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,7 +141,7 @@ tox-uv = "tox_uv"
 ty = "ty"
 
 [tool.pytest.ini_options]
-addopts = "-s -v -x --strict-markers -m 'not extra' --doctest-modules --cov=src"
+addopts = "-s -v -x --strict-markers -m 'not extra and not network' --doctest-modules --cov=src"
 doctest_optionflags = [
   "ALLOW_BYTES",
   "ALLOW_UNICODE",
@@ -150,7 +150,7 @@ doctest_optionflags = [
   "NORMALIZE_WHITESPACE"
 ]
 markers = [
-  "network: tests that require network connectivity",
+  "network: uncachable tests that always require network connectivity",
   "slow: slow tests that should be run infrequently",
 
   "extra",
--- a/tests/test_hgvs_dataproviders_uta.py
+++ b/tests/test_hgvs_dataproviders_uta.py
@@ -127,6 +127,7 @@ class Test_hgvs_dataproviders_uta_UTA_default_with_pooling(unittest.TestCase, UT
         )
 
 
+@pytest.mark.network
 class Test_hgvs_dataproviders_uta_with_pooling_without_cache(unittest.TestCase, UTA_Base):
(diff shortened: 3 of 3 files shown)


reece:
@andreasprlic Remarkably easy. I wish I'd done this years ago. I'd appreciate a quick review

tests/conftest.py
```diff
@@ -30,14 +31,13 @@ def vcr_config(request):  # noqa: ARG001
     }
 
 
-hgvs.easy.hdp = hgvs.dataproviders.uta.connect(
-    mode=os.environ.get("HGVS_CACHE_MODE", "run"), cache=CACHE
-)
+_hdp = hgvs.dataproviders.uta.connect(mode=os.environ.get("HGVS_CACHE_MODE", "run"), cache=CACHE)
+_parser = Parser()
```
andreasprlic, line 35: Nit and prob not an issue. I believe this declaration makes this a class instance and not just a session instance the way the fixtures below are indicating.

andreasprlic (approved):
LGTM

Issue #871, opened 2026-08-09
Replace literal string/int tokens with enums (var.type, edit.type, shuffle_direction, etc.)


reece:
## Problem

Several concepts throughout the codebase are represented as bare string or int literals compared directly, rather than enum members. This showed up concretely while working #850: `Normalizer`/`DataCompiler`'s `shuffle_direction`/`direction` params compare against literal `3`/`5` (meaning 3' and 5' shuffle) in ~9 sites across `normalizer.py`/`datacompiler.py`, duplicating a concept `variantmapper.py` already has as private `_SHUFFLE_3PRIME`/`_SHUFFLE_5PRIME` constants that nothing else uses. That was deliberately left as documented per-file-ignore debt (see #850) rather than fixed piecemeal, with a note that centralizing was "a reasonable follow-up."

The same pattern is much larger than just shuffle direction:

- `SequenceVariant.type` (and `var.type`) is compared against the single-letter literals `"c"`, `"g"`, `"m"`, `"n"`, `"p"`, `"r"` at ~46 sites across `src/hgvs/`.
- `Edit.type` is compared against `"sub"`, `"del"`, `"ins"`, `"dup"`, `"delins"`, `"inv"`, `"identity"`, `"con"` at ~36 sites.
- `Normalizer.shuffle_direction`/`DataCompiler.get_shuffled_variant`'s `direction` param: literal `3`/`5` (HGVS 3'/5' convention) at ~9 sites, plus the `"LEFTSHUFFLE"`/`"RIGHTSHUFFLE"`/`"EXPAND"` mode strings passed to `bioutils.normalize` in `datacompiler.py`.

None of these are currently typed as enums, so there's no autocomplete, no exhaustiveness checking, and typos (`"delins"` vs `"del_ins"`, `3` vs `5`) fail silently or produce wrong behavior rather than a clear error.

## Scope

1. Introduce enums (or reuse/extend `hgvs.enums`) for:
   - Variant type (`c`/`g`/`m`/`n`/`p`/`r`) -- likely the highest-value change given 46 call sites, but also the highest-risk since `SequenceVariant.type` is public API (serialized in `__str__`/`__repr__`, likely compared as plain strings by downstream consumers too -- needs a compatibility plan, e.g. a `StrEnum` that still equals its string value).
   - Edit type (`sub`/`del`/`ins`/`dup`/`delins`/`inv`/`identity`/`con`).
[...]

Issue #869, opened 2026-08-09
Reduce function complexity flagged by PLR0911/PLR0912/PLR0915 (per-file-ignores exceptions)


reece:
## Problem

#850 is paying down the frozen ruff lint debt from #838's `per-file-ignores`, rule-by-rule. The `PLR09xx` complexity rules (too-many-return-statements, too-many-branches, too-many-statements) are the largest remaining chunk and are qualitatively different from the rest: fixing them means refactoring function bodies (splitting logic, extracting helpers), not mechanical rewrites. This issue splits that work out so #850 can close on the mechanical rules without blocking on refactor design decisions.

## Scope

Every file/rule pair below is currently listed under `[tool.ruff.lint.per-file-ignores]` in `pyproject.toml` for `PLR0911` (too many return statements), `PLR0912` (too many branches), and/or `PLR0915` (too many statements):

- `src/hgvs/alignmentmapper.py` -- PLR0911, PLR0912, PLR0915
- `src/hgvs/decorators/lru_cache.py` -- PLR0915 (note: this file's other ignores, A002/B006/B008/N806, are a documented permanent exception -- a verbatim CPython lru_cache backport -- but PLR0915 itself wasn't part of that exception and is still open)
- `src/hgvs/edit.py` -- PLR0912
- `src/hgvs/location.py` -- PLR0911, PLR0912
- `src/hgvs/normalizer.py` -- PLR0912, PLR0915
- `src/hgvs/pretty/console/protseqrenderer.py` -- PLR0912, PLR0915
- `src/hgvs/pretty/console/shuffledvariant.py` -- PLR0912, PLR0915
- `src/hgvs/pretty/console/txaligrenderer.py` -- PLR0912, PLR0915
- `src/hgvs/pretty/datacompiler.py` -- PLR0912, PLR0915
- `src/hgvs/pretty/prettyprint.py` -- PLR0912, PLR0915
- `src/hgvs/repeats.py` -- PLR0912, PLR0915
- `src/hgvs/utils/altseq_to_hgvsp.py` -- PLR0912, PLR0915
- `src/hgvs/utils/position.py` -- PLR0911, PLR0912
- `src/hgvs/variantmapper.py` -- PLR0912, PLR0915

(non-exhaustive if the list has drifted since this was filed -- see current `pyproject.toml` for the authoritative list.)

[...]

Issue #849, opened 2026-08-08
Avoid database instantiation during testing


reece:
An hdp connects to a postgresql database on instantiation, *even when in testing mode* where (we hope) the data are entirely cached. This creates several headaches. 

First, it means that the databsae must be accessible from github action environments essentially full-time even though it won't be used.

Second, because the database connection is part of the cache key, it means that the cache must be built using that environment, which makes it much harder to use localhost data during development.

There are several solutions, but it mostly signals that the hdp setup needs serious rethinking for 2.0.

reece:
Making this the parent/tracking issue for hdp-testing devops pain, based on recurring friction observed across today's session (2026-08-08):

- **Binary cache merge conflicts**: `tests/data/cache-py3.hdp` is a pickled binary blob. Every branch that touches cached data (fix/801, issue_505_replace_parsley, fix/714-splice-region-preserved, generator_exit_error — all today) hit manual, error-prone conflict resolution on this file. Hand-resolution has already let schema-mismatched rows slip through uncaught (e.g. `get_tx_exons` rows for NM_182758.2/NM_000267.3 during the psycopg2→psycopg list→dict migration) — no diffability, no validation on merge. This is a strong argument for #450's "portable cache file format" goal, and probably wants a human/diffable format (e.g. JSON/JSONL keyed and sorted) instead of pickle.
- Related: #450 (untangle lru_cache/persistent cache — includes "portable cache file format" and "enable caches to be used as sole-source for data for testing", directly relevant here), #858 (ensure tests use fixtures), #859 (schema drift / connection hang against live UTA), #867 (silent fallback to live UTA on cache miss).

Filing these as related rather than separately, per discussion — this issue is the parent for the broader hdp/testing rethink.

Issue #838, opened 2026-08-07
Reformat all branches consistently to simplify PR diffs


reece:
## Problem

We adopted ruff/pre-commit but never applied it to all files. Now pre-commit reformats whatever files a branch touches, so formatting changes get comingled with logic changes in every PR diff. Reformatting everything at once on `main` also surfaces a large number of pre-existing lint failures that aren't related to any one branch's changes.

## Plan

1. Run `uvx ruff format src` on `main`. Add file-specific `pyproject.toml` ignores for every rule still failing post-format, so `pre-commit run --all-files` passes clean. Commit that as a single "reformat + freeze existing lint debt" commit on `main`. (This freezes existing lint debt as accepted-for-now, not fixed.)
2. Branch owners: on your branch, **before** merging main, run `uvx ruff format src` and commit that as its own "reformat" commit using `git commit --no-verify`. (Pre-commit will still fail here since your branch doesn't have main's lint-debt ignores yet — the formatting itself is correct, so `--no-verify` is safe for this one commit.) Then merge `main` in. Since both sides are now formatted identically, git resolves the formatting-only hunks cleanly, only genuine content conflicts remain to fix by hand, and you pick up main's ignores so pre-commit passes normally from then on. (Reformatting *after* merging main instead reintroduces the comingled-diff problem this issue exists to fix — always reformat first, merge second.) **Don't format outside of src for now.**
3. Review and merge PRs as usual (but faster without the irrelevant diffs)

**Phase 2** (separate, later): address the frozen lint debt (TRY003, EM102, S101, SIM108, etc.) deliberately on `main`, not blocking any branch merges.

## Branches / owners

- @andreasprlic — #836, #835, #821
- @igor-petrik-invitae — #831
- @davmlaw — #826
- @b0d0nne11 — #719

Once step 1 lands on `main`, please do step 2 on your branch(es) above.

reece:
Subsumes #803 #832

reece:
This is substantially done. #831 has been updated but not merged, pending discussion. The rest have been merged. 
Next step is #850.

PR #872, opened 2026-08-09
chore: pay down frozen ruff lint debt (per-file-ignores), rule by rule

### LINKED ISSUE #850: Clean up frozen lint/formatting debt (per-file-ignores exceptions)
## Problem

#838 froze a large set of pre-existing ruff violations behind `[tool.ruff.lint.per-file-ignores]` in `pyproject.toml` so `main` and branches could reformat without comingling formatting and logic diffs. That issue's own "Phase 2" names the follow-up explicitly:

> address the frozen lint debt (TRY003, EM102, S101, SIM108, etc.) deliberately on `main`, not blocking any branch merges.

That phase was never tracked as its own issue. This issue is that tracking.

## Scope

Every file/rule pair currently listed under `[tool.ruff.lint.per-file-ignores]` in `pyproject.toml` is debt to pay down, e.g. (non-exhaustive, see current `pyproject.toml` for the authoritative list):

- `tests/*` = `PLR2004`, `S101` (assert usage in tests, magic values)
- `src/hgvs/location.py`, `src/hgvs/normalizer.py`, `src/hgvs/pretty/*` — `TRY003`/`EM101`/`EM102` (long exception messages), `SIM102`/`SIM108` (collapsible/ternary opportunities)
- `src/hgvs/dataproviders/{ncbi,uta}.py` — `PTH119`, `RUF012`, `UP008`, `TRY002`/`TRY003`
- `src/hgvs/utils/PersistentDict.py` — `N999`, `S301` (pickle usage), `PTH123`

Additionally, `sbin/generate_parser.py` currently fails ruff outright (`PTH110`/`PTH118`/`PTH120`/`PTH123`, `S324` insecure `hashlib.md5`, `SIM115`) with **no** per-file-ignore entry — it was missed when #838 froze debt for `src/` and isn't covered since that pass deliberately excluded `sbin/`. This should either be fixed outright or get its own ignore entry as part of this cleanup.

**Out of scope:** the `PLR09xx` complexity rules (too many branches/statements/returns) are tracked separately in #869, since fixing those means refactoring function bodies rather than mechanical rewrites.

## Suggested approach

1. Go rule-by-rule (or file-by-file) through the frozen ignores, fixing violations and shrinking/removing the corresponding `per-file-ignores` entry.
2. Prioritize security/correctness-flavored rules first (`S101`, `S301`, `S324`) over style rules (`SIM*`, `UP*`).
3. Land in small, focused PRs (one rule or one file-group at a time) to keep review tractable, per this repo's usual single-issue-per-PR convention.
4. Once `per-file-ignores` is empty (or down to genuinely permanent exceptions), remove the section and this issue closes.

## Related

- #838 (predecessor — froze this debt so branches could reformat cleanly)
- #869 (split out — `PLR09xx` function-complexity refactors)

reece:
## Goal

Work through #850: pay down the ruff lint debt that #838 froze behind `[tool.ruff.lint.per-file-ignores]`, one rule/file group at a time, security/correctness rules first, without mixing formatting and logic diffs.

## Work done

Fixed or converted to a real exception, with tests passing throughout (443 → 420 as the suite's own network-isolation work in #870 landed on this branch):

- `sbin/generate_parser.py`: fixed outright (was failing ruff with no ignore entry at all) — pathlib, `hashlib.md5(usedforsecurity=False)`, context-managed opens.
- Mechanical style/logging/builtin-shadowing/pathlib/simplify rules: `UP031/UP036/UP030/UP008`, `C401/C408`, `B028`, `PT018`, `PLR0402/RUF046`, `G001/G003/G004`, `A001/A002`, `PTH119/PTH123`, `RET503`, `SIM102/SIM108`.
- Exception style: `TRY002/TRY003/EM101/EM102` (extracted messages to locals) — found and fixed a real latent bug along the way (`Normalizer`'s "conversion variants" error passed a stray positional arg instead of formatting the message).
- Bugbear: `B904` (exception chaining), `B026` (dropped a dead `*args` param).
- Pylint comparison/branch: `PLR1714` (merged repeated equality checks), `PLR5501` (flattened `else: if` — including a non-trivial reindent in `altseq_to_hgvsp.py`, verified against coverage), `PLR1730`, `PLR1704`, `RUF012/RUF013`.
- `ARG002/ARG004`: fixed the one genuinely-dead parameter (`datacompiler.py`'s `tx_exons`); the other 8 are documented permanent exceptions (interface/polymorphism constraints ruff can't see across files).
- `PLC0415`: moved 8 non-lazy local imports to module level; documented 4 genuinely lazy ones (optional deps, backend dispatch, deferred DB connection).
- `PLR2004`: resolved or documented every site (HGVS 3'/5' convention, version-parsing helper, display heuristics) — see "Remaining work" below for the follow-up this produced.
- **`S101` (the security-priority item)**: traced every non-test `assert`'s call sites to classify EXTERNAL-INPUT (validates caller-supplied values -- stripped under `-O`, a real correctness gap) vs INTERNAL-INVARIANT (only reachable via this library's own logic). Converted 17 external-input sites to `TypeError`/`ValueError`/`HGVSError` as appropriate; documented 9 internal-invariant sites as permanent exceptions. Updated two tests that were pinned to the old accidental `AssertionError` rather than a designed contract.
- Scrutinized the whole `per-file-ignores` block for staleness: found and fixed one of my own false negatives (`TRY300`, which I'd misread as resolved while its ignore was still masking it), removed 2 confirmed-stale `PLR0915` entries, fixed a missing doc-comment.
- `Makefile`: added `make ruff-counts`, a read-only reporting target (violations by file/code with `per-file-ignores` disabled) for tracking what's left.

## Remaining work

Rule codes still present in `per-file-ignores` (all now either genuinely permanent, or explicitly tracked elsewhere -- none are silently deferred):

- **`PLR0911`/`PLR0912`/`PLR0915`** (function complexity, 13 files) -- split out to **#869**, since fixing these means refactoring function bodies, not mechanical rewrites.
- **`S301`** (pickle in `PersistentDict.py`) -- blocked on **#849**'s planned rethink of `hdp` caching for 2.0; reworking it now would be throwaway.
- **`N999`** (module naming in `PersistentDict.py`) -- same #849 blocker.
- **`A002`/`B006`/`B008`/`N806`** (`lru_cache.py`) -- permanent exception, verbatim CPython `lru_cache` backport recipe.
- **`N805`** (`location.py`) -- permanent exception, deliberate `lhs`/`rhs` naming for symmetric comparison dunders.
- **`PLW1641`** (`location.py`) -- permanent exception, adding `__hash__` to a mutable value type is a correctness hazard, not a lint fix.
- **`PT009`/`PT027`** (`tests/*`, 446 sites across 23 files) -- pre-existing `unittest`-style debt, unrelated to #850, surfaced only because this branch happened to touch one of those files; too large to fix incidentally.
- **`PLR2004`** (`normalizer.py`/`datacompiler.py`'s shuffle-direction `3`/`5`, plus `interface.py`/`norm.py`/`context.py`) -- documented as permanent in-place, but centralizing the shuffle-direction magic values into a shared enum was flagged as worth doing properly -- filed as **#871**.

New issues filed this session:
- **#869** -- Reduce function complexity flagged by PLR0911/PLR0912/PLR0915 (split from #850).
- **#871** -- Replace literal string/int tokens with enums (`var.type`, `edit.type`, `shuffle_direction`, etc.), including revisiting whether `hgvs.utils.orderedenum.OrderedEnum` is still needed now that Python 3.11+'s `IntEnum` has native ordering.

Also fixes #850's originally-scoped gap in `sbin/generate_parser.py`; #850 itself stays open pending #869/#871 and the `S301`/#849 blocker.

## Test plan

- [x] Full test suite passes (420 passed, 44 skipped, 28 deselected, 4 xfailed)
- [x] `ruff check`/`ruff format` clean except the pre-existing, out-of-scope preview-rule violations (`A003`/`PLR0917`/`RUF059` -- not in `per-file-ignores`, present on `main` before this branch, unrelated to #850)
- [x] `pyproject.toml` validated as parseable TOML after every edit
- [x] Every per-file-ignore removal double-checked by isolating that specific rule/file pair and re-running ruff with the ignore actually absent (not just trusting a stale check)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


jsstevenson (approved):
I gave this a 5 minute lookover, so not an exhaustive review; but everything looks good to me

PR #882, opened 2026-08-28
Fix/881 prettyprint attributeerror start

### LINKED ISSUE #881: PrettyPrint.display() crashes with AttributeError: 'X' object has no attribute 'start' for intronic/offset and dup variants
### Summary

`PrettyPrint.display()` (and, more generally, `AlignmentMapper.n_to_c()`) raises `AttributeError: 'BaseOffsetPosition' object has no attribute 'start'` for an intronic/offset (`c.N+M` / `c.N-M`) variant whose display window includes a transcript/genome alignment gap. A closely related `AttributeError: 'SimplePosition' object has no attribute 'start'` also occurs for any `dup` variant passed to `PrettyPrint.display()`.

### Environment

- hgvs version: `2.0.0a1.dev183+gf1a39c4e6` (main @ `f1a39c4`)
- Python: 3.12

### Steps to reproduce (offset/intronic crash)

```python
import hgvs.dataproviders.uta
from hgvs.assemblymapper import AssemblyMapper
from hgvs.parsers import Parser
from hgvs.pretty.prettyprint import PrettyPrint

hdp = hgvs.dataproviders.uta.connect(mode=None, cache=None)
hp = Parser()
am = AssemblyMapper(hdp, assembly_name="GRCh37")
pp = PrettyPrint(hdp, am, reverse_display=False)

var_c = hp.parse("NM_000682.6:c.901+5G>A")
pp.display(var_c)
```

**Traceback:**
```
Traceback (most recent call last):
  ...
  File "hgvs/pretty/datacompiler.py", line 318, in data
    self._backfill_gap_in_ref(...)
  File "hgvs/pretty/datacompiler.py", line 448, in _backfill_gap_in_ref
    c_interval = mapper.n_to_c(n_interval, strict_bounds=False)
  File "hgvs/alignmentmapper.py", line 567, in n_to_c
    start = n_interval.start.start.base
AttributeError: 'BaseOffsetPosition' object has no attribute 'start'
```

### Steps to reproduce (dup crash)

```python
var_g = hp.parse("NC_000007.13:g.36561663dup")
pp.display(var_g, "NM_001177507.2")
```

**Traceback:**
```
  File "hgvs/pretty/datacompiler.py", line 134, in get_position_and_state
    start, end = self._get_start_end(sv)
  File "hgvs/pretty/datacompiler.py", line 81, in _get_start_end
    s = var.posedit.pos.start.start
AttributeError: 'SimplePosition' object has no attribute 'start'
```

### Root cause

Both crashes are the same defect pattern: code branches on the type of the *outer* interval object instead of the type of the specific `.start`/`.end` attribute it's about to dereference, so it wrongly assumes `.start` is itself a nested `Interval`:

- `hgvs/pretty/datacompiler.py`, `DataCompiler._get_start_end()`: `elif isinstance(var.posedit.pos, hgvs.location.Interval): s = var.posedit.pos.start.start` — should check `isinstance(var.posedit.pos.start, hgvs.location.Interval)`, matching the (correct) sibling branch used two lines later for `.end`.
- `hgvs/pretty/datacompiler.py`, `DataCompiler._backfill_gap_in_ref()`: builds a plain `hgvs.location.Interval` of two `BaseOffsetPosition`s and passes it to `AlignmentMapper.n_to_c()`, which only knows how to unwrap a `BaseOffsetInterval` (`.start.base`) or an `Interval`-of-`Interval`s (`.start.start.base`) — a plain `Interval` of `BaseOffsetPosition`s matches neither.
- `hgvs/alignmentmapper.py`, `AlignmentMapper.n_to_c()`: the preliminary bounds-check block has the same outer-vs-inner type-check bug (`elif isinstance(n_interval, Interval): start = n_interval.start.start.base`), inconsistent with the correct nested-interval handling later in the same method (`if isinstance(n_interval.start, Interval): ...`).

### Expected behavior

`PrettyPrint.display()` should render the variant (including intronic/offset positions and `dup` edits) without raising.

andreasprlic:
Fix PrettyPrint AttributeError on intronic/offset and dup variants

## Summary

Fixes #881.

Several `PrettyPrint`/`DataCompiler`/`AlignmentMapper` code paths branched
on the type of an *outer* `Interval` instead of the type of the specific
`.start`/`.end` attribute they were about to dereference, then wrongly
assumed `.start` was itself a nested `Interval`:

- `DataCompiler._get_start_end()` misclassified plain (non-`BaseOffset`)
  positions for `dup` variants, crashing with
  `'SimplePosition' object has no attribute 'start'`.
- `DataCompiler._backfill_gap_in_ref()` built a plain `Interval` (instead
  of `BaseOffsetInterval`) when back-filling alignment gaps near
  intronic/offset positions, crashing `AlignmentMapper.n_to_c()` with
  `'BaseOffsetPosition' object has no attribute 'start'`.
- `AlignmentMapper.n_to_c()` had the same latent bug in its bounds-check
  preamble, reachable by any caller passing a plain `Interval` per its
  own type hint — fixed defensively even though no in-repo caller hits it
  after the above fix.

Fixing the `dup` branch also surfaced a second, independent bug:
`get_position_and_state()` returned raw `Position` objects instead of
`int`s for `dup` variants, which `bioutils.normalize()` rejects
downstream with a `TypeError`.

## Test plan

- [x] New regression test reproducing the intronic/offset crash
      (`Test_SimplePosition.test_intronic_offset_near_alignment_gap`)
- [x] New offline, CI-enabled regression test for the `dup` crash
      (`Test_DataCompilerDup.test_get_position_and_state_dup`), targeting
      `DataCompiler.get_position_and_state()` directly so it doesn't need
      the full transcript/alignment pipeline
- [x] New regression test for the `AlignmentMapper.n_to_c()` defensive fix
      (`test_n_to_c_plain_interval_of_base_offset_positions`)
- [x] Confirmed each new test fails against the pre-fix code and passes
      after
- [x] Full live-network `PrettyPrint` test class: crashes went from 4 to 0
      (2 unrelated, pre-existing content-assertion failures remain, not
      part of this bug)
- [x] `uv run ruff check` / `ruff format --check` / `mypy` — no new
      findings
- [x] `uv run pytest -m quick` — all passing


src/hgvs/alignmentmapper.py
```diff
@@ -563,9 +563,12 @@ def n_to_c(
         if isinstance(n_interval, BaseOffsetInterval):
             start = n_interval.start.base
             end = n_interval.end.base
-        elif isinstance(n_interval, Interval):
+        elif isinstance(n_interval.start, Interval):
             start = n_interval.start.start.base
             end = n_interval.end.end.base
+        else:
+            start = n_interval.start.base
+            end = n_interval.end.base
```
reece, line 571: @andreasprlic See SonarCloud analysis. It correctly identies that the first and else branches are identical.

reece (changes requested):

